PR #1, opened 2024-06-22
Restore config before locking (and writing config)

surfdado:


lukash:
Can you add a changelog record (with the `"Fix: ..."` git trailer)? Thanks!

PR #5, opened 2024-07-25
LEDs: add felony led effect

acheronfail:
My kids found this effect amusing from the old `float` package, so I ported it over to this one, too!

### DIFF
--- a/src/leds.c
+++ b/src/leds.c
@@ -169,12 +169,25 @@ static void led_set_color(
     leds->led_data[led] = RGBW(r, g, b, w);
 }
 
+static void strip_set_color_range(
+    Leds *leds,
+    const LedStrip *strip,
+    uint32_t color,
+    float brightness,
+    float blend,
+    uint8_t idx_start,
+    uint8_t idx_end
+) {
+    uint8_t end = idx_end < strip->length ? idx_end : strip->length;
+    for (uint8_t i = idx_start; i < end; ++i) {
+        led_set_color(leds, strip, i, color, brightness, blend);
+    }
+}
+
 static void strip_set_color(
     Leds *leds, const LedStrip *strip, uint32_t color, float brightness, float blend
 ) {
-    for (uint8_t i = 0; i < strip->length; ++i) {
-        led_set_color(leds, strip, i, color, brightness, blend);
-    }
+    strip_set_color_range(leds, strip, color, brightness, blend, 0, strip->length);
 }
 
 // Returns a cosine wave oscillating from 0 to 1, starting at 0, with a period of 2s.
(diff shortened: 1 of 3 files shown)


src/leds.c
```diff
@@ -261,6 +275,33 @@ static void anim_knight_rider(Leds *leds, const LedStrip *strip, const LedBar *b
     }
 }
 
+static void anim_felony(Leds *leds, const LedStrip *strip, float time) {
+    uint32_t color_off = 0x00000000;
+    uint32_t color_red = 0x00ff0000;
+    uint32_t color_blu = 0x000000ff;
```
lukash: Could you take the two colors from the configuration, so that the effect can be configured in any color combination?

lukash (commented):
Hey, thanks for the PR, I was looking forward to people contributing effects, and nicely written, too.

src/leds.c
acheronfail: Ooh, that's a great idea! Done :+1:

lukash:
Apologies for the long turnaround.

I wanted to test this, and then I also thought about LED strips with an odd number of LEDs. The way it is one half of the effect will be longer. What about making the center LED always black to make it symmetric?

Also, feel free to squash the second commit to the first and when making changes just amend the outstanding commit, I don't need to see the differences in separate ones.

Lastly, would you add a line like this to your commit message, it's a git trailer and will ensure automatic changelog item (with an attribution to you):
```
Feature: Add the Felony LED effect.
```

Thank you!

acheronfail:
> I wanted to test this, and then I also thought about LED strips with an odd number of LEDs. The way it is one half of the effect will be longer. What about making the center LED always black to make it symmetric?

I have some test footage I can upload here if you'd like. Unfortunately I only have LED strips with even numbers of LEDs, so I can't test out odd variations. But I did like your suggestion, so I implemented it.

> Also, feel free to squash the second commit to the first and when making changes just amend the outstanding commit, I don't need to see the differences in separate ones.
> 
> Lastly, would you add a line like this to your commit message [...]

Squashed, amended commit message with trailer and rebased on the latest trunk. :slightly_smiling_face:

src/leds.c
```diff
@@ -261,6 +275,40 @@ static void anim_knight_rider(Leds *leds, const LedStrip *strip, const LedBar *b
     }
 }
 
+static void anim_felony(Leds *leds, const LedStrip *strip, const LedBar *bar, float time) {
+    uint32_t color_off = 0x00000000;
+
+    float state_duration = 0.05f;
```
lukash: Please use `static const` for the above two constants.

src/leds.c
```diff
@@ -169,14 +169,28 @@ static void led_set_color(
...
+    Leds *leds,
+    const LedStrip *strip,
+    uint32_t color,
+    float brightness,
+    float blend,
+    uint8_t idx_start,
+    uint8_t idx_end
 ) {
-    for (uint8_t i = 0; i < strip->length; ++i) {
+    uint8_t start = idx_start < strip->length ? idx_start : 0;
```
lukash: I don't think you need this calculation, if `start` would be greater than `end`, the for loop below just won't execute at all (programmer error).

src/leds.c
```diff
@@ -261,6 +275,40 @@ static void anim_knight_rider(Leds *leds, const LedStrip *strip, const LedBar *b
...
 
+static void anim_felony(Leds *leds, const LedStrip *strip, const LedBar *bar, float time) {
+    uint32_t color_off = 0x00000000;
+
+    float state_duration = 0.05f;
+    float state_mod = fmodf(time, 3.0f * state_duration);
+
+    // also account for led strips with odd numbers of leds (leaving the middle one black)
+    uint8_t stop_idx = strip->length / 2;
+    uint8_t start_idx = strip->length / 2 + (strip->length % 2 * 2);
```
lukash: I think the `* 2` is wrong, it'll result in two black LEDs unless I'm mistaken.

lukash (commented):
> I only have LED strips with even numbers of LEDs, so I can't test out odd variations.

You can test it easily, just configure the strip with one less LED :grinning: 

I'll try to test it myself tonight though.

src/leds.c
acheronfail: Done!

src/leds.c
acheronfail: Good observation, also done!

src/leds.c
acheronfail: Ahh, the 3rd hardest of the 2 hardest things in programming :man_facepalming: fixed!

acheronfail:
> You can test it easily, just configure the strip with one less LED 😀

Of course :man_facepalming: 

I tested it (at a slow speed `0.2`) with the strip set to 20 and 19 leds, as well as with different colours. You can see the results in these videos here:

https://github.com/user-attachments/assets/4eb50c5e-eefe-4444-9a79-f3df6f32af25

https://github.com/user-attachments/assets/9476b68c-815f-4f6f-ba9a-e00bfe6fe08f

lukash:
Thank you!

PR #7, opened 2024-07-30
New brake current behavior for 6.05 firmware

surfdado:
Compatible with 6.02 as well

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1052,7 +1052,22 @@ static void brake(data *d) {
     }
 
     VESC_IF->timeout_reset();
-    VESC_IF->mc_set_brake_current(d->float_conf.brake_current);
+
+    // If brake current is set to 0 don't do anything
+    if (d->float_conf.brake_current == 0) {
+        return;
+    }
+
+    // Use brake current over certain ERPM to prevent the board skidding to a stop when deactivated
+    // at speed?
+    if (d->motor.abs_erpm > 2000) {
+        VESC_IF->mc_set_brake_current(d->float_conf.brake_current);
+        return;
+    }
+
+    // Use DC control mode as it has better holding power
+    // Also improves with 6.05 shorting feature
+    VESC_IF->mc_set_duty(0);
 }
 
 static void set_current(data *d, float current) {


src/main.c
```diff
@@ -1052,7 +1052,22 @@ static void brake(data *d) {
...
     VESC_IF->timeout_reset();
-    VESC_IF->mc_set_brake_current(d->float_conf.brake_current);
+
+    // If brake current is set to 0 don't do anything
+    if (d->float_conf.brake_current == 0) {
+        return;
+    }
+
+    // Use brake current over certain ERPM to prevent the board skidding to a stop when deactivated
+    // at speed?
```
lukash, line 1062: Is the question mark in the comment intentional? It doesn't make me feel very confident :slightly_smiling_face:

src/main.c
```diff
@@ -1052,7 +1052,22 @@ static void brake(data *d) {
...
+    if (d->motor.abs_erpm > 2000) {
+        VESC_IF->mc_set_brake_current(d->float_conf.brake_current);
+        return;
+    }
+
+    // Use DC control mode as it has better holding power
+    // Also improves with 6.05 shorting feature
+    VESC_IF->mc_set_duty(0);
```
lukash, line 1070: I wonder if you'd always want to do this, not just when movement was detected (on line 1046 above) and it didn't time out?

surfdado:
Sorry, the comment was there before I tested it, should be removed

src/main.c
surfdado: > I wonder if you'd always want to do this, not just when movement was detected (on line 1046 above) and it didn't time out?

that would be a question for the VESC Discord. My guess is that it would waste energy and drain our batteries. That's why this entire logic exists (we used to not have it and brake current was keeping our ESCs warm...)

techfoundrynz:
> Sorry, the comment was there before I tested it, should be removed

Sorry about that comment. Made sense when pitching the idea to you guys.

src/main.c
lukash: No response on the discord, I guess I'll do some testing when I have a chance...

Relys:
We should also change https://github.com/lukash/refloat/blob/851431a4cd35e93e1630e0f1f65627922a4a0388/src/main.c#L1046

`if (d->motor.abs_erpm > 5 || d->brake_timeout == 0) {`

as discussed https://discordapp.com/channels/904830990319485030/1092899715273740470/1277394682166378609

Jared-Is-Coding:
I've been riding this for the last two weeks along with @lukash's [Sticky Brake changes](https://github.com/lukash/refloat/commit/5fa87b66c85dfd7b459ea42d31be6eb5c3bb0776) and have found it to work very well.

No complaints, seems to do a great job.

lukash:
Thank you, closing in favor of a slightly different implementation: https://github.com/lukash/refloat/pull/22

PR #8, opened 2024-07-30
Add "make clang" target

surfdado:


Makefile
```diff
@@ -19,6 +19,9 @@ else
     MINIFY_CMD="cat"
 endif
 
+clang:
+	clang-format-17 -i src/*.c src/*.h
```
lukash, line 23: This is insufficient, as you can have source files in subdirectories. I don't know of an easy way to do this in a Makefile (though there may be one, likely needing to write a function or something), that's why the target isn't here. I use `lefthook` not only for git hooks, but also to manually run the command.

Also, naming the target `clang` is misleading, `clang` is the compiler, this should be named `clang-format`.

surfdado:
No worries, I'll just keep this local then

PR #9, opened 2024-08-02
Feature: Port Haptic Feedback over from Float Package

Relys:
Pretty straight forward. Simply added Haptic Feedback from float package which is a widely requested feature.

Future development will be supporting the new 6.05 foc play tone feature:
//e.g.
//if ((d->fw_version_major + d->fw_version_minor * 0.01) >= 6.05)
//VESC_IF->foc_play_tone(int channel, float freq, float voltage);
//VESC_IF->foc_play_tone(0, 740.0, 3.0);
//VESC_IF->foc_stop_audio(bool reset);
//VESC_IF->foc_stop_audio(true);

I also plan on submitting a PR for adding bms tiltback/haptic feedback along with a new command for the BMS to call refloat package and update information about the BMS state. BMS state and faults will be calculated in lisp to pull temp/cell data.


Relys:
Refactored into its own file.

.gitignore
```diff
@@ -18,3 +18,4 @@ confxml.h
 *.obj
 *.so
 *.rcc
+.vscode/settings.json
```
lukash: Unrelated to the PR and IDE ignores should be put into your local `gitignore` (e.g. `~/.config/git/ignore`).

src/haptic_buzz.c
```diff
@@ -0,0 +1,123 @@
...
+    haptic_buzz->applied_haptic_current = 0;
+}
+
+float haptic_buzz_update(
+    HapticBuzz *haptic_buzz,
+    State *state,
+    float current_time,
+    float abs_erpm,
+    float startup_click_current,
+    float note_period,
```
lukash: This should probably rather be called `note_duration` or `note_length`, using "period" is confusing.

src/haptic_buzz.c
```diff
@@ -0,0 +1,123 @@
...
+    bool brake
```
lukash: I think this should be named `disengage` rather than `brake`.

src/haptic_buzz.c
```diff
@@ -0,0 +1,123 @@
...
+                    if (haptic_buzz->haptic_type == HAPTIC_BUZZ_ALTERNATING) {
+                        haptic_buzz->haptic_mode = 5 - haptic_buzz->haptic_mode;
+                    } else {
+                        haptic_buzz->haptic_mode = 1 - haptic_buzz->haptic_mode;
+                    }
+                }
+
+                haptic_buzz->haptic_timer = current_time;
+            }
+        }
```
lukash: IMO this piece of code should be factored out into an API similar to that of the `foc_play_tone()`, that is e.g. `buzz_tone(float frequency, float current)`. It'll be required to add the 6.05 `foc_play_tone()` as an alternative in a clean way.

Then, `haptic_buzz_update()` will prepare the data for the particular tone to play (according to the type of buzz) into the `hapitc_buzz` structure and `buzz_tone()` can modulate the buzz current onto the output current.

src/haptic_buzz.c
```diff
@@ -0,0 +1,123 @@
...
+    haptic_buzz->haptic_buzz_hv = float_conf->haptic_buzz_hv;
+    haptic_buzz->haptic_buzz_lv = float_conf->haptic_buzz_lv;
+    haptic_buzz->haptic_buzz_temp = float_conf->haptic_buzz_temp;
+}
+
+void haptic_buzz_reset(HapticBuzz *haptic_buzz, float current_time) {
+    haptic_buzz->haptic_tone_in_progress = false;
+    haptic_buzz->haptic_timer = current_time;
+    haptic_buzz->applied_haptic_current = 0;
+}
```
lukash: The function doesn't completely initialize the structure. The remaining variables are initialized on lines 117-120, relying on no buzz being played on the first call. Please make the initialization here complete for clarity and simplicity.

lukash (changes requested):
Thanks for the PR! Pretty clean as a straightforward port, but I'd like to streamline the implementation a bit more. I think adding the support for `foc_play_tone()` to this PR will help fleshing out the code to its final state. What I'd do is change the current commit to allow for easy integration of `foc_play_tone()` (see my comments) and then add a small commit that puts it in place.

For the 6.05 API check out the [`dev/6.05`](https://github.com/lukash/refloat/tree/dev/6.05) branch.

.gitignore
Relys: This should be reverted now.

src/haptic_buzz.c
Relys: Renamed parameter to note_duration.

src/haptic_buzz.c
Relys: I agree, I was confused too when porting it.

src/haptic_buzz.c
Relys: Done

src/haptic_buzz.c
Relys: Not exactly sure how to approach this. First of all IIRC haptic_tone is limited to specific frequency range and cannot play arbitrary frequencies @surfdado Also, with the current implementation, haptic tone stops playing when haptic_feedback_update isn't called in refloat_thd (as it's not being modulated onto the current anymore). This is different than foc_play_tone() and foc_stop_audio() which will start playing the tone once foc_play_tone() is called and foc_stop_audio() needs to be explicitly called to stop playing. i.e. That means the both methods have uniquely different requirements regarding state management in the finite state automata defined in refloat_thd.

lukash: > First of all IIRC haptic_tone is limited to specific frequency range and cannot play arbitrary frequencies

Oh, right, so maybe the vibrating haptics will have to remain via the current modulation?

> Also, with the current implementation, haptic tone stops playing when haptic_feedback_update isn't called in refloat_thd (as it's not being modulated onto the current anymore). This is different than foc_play_tone() and foc_stop_audio() which will start playing the tone once foc_play_tone() is called and foc_stop_audio() needs to be explicitly called to stop playing. i.e. That means the both methods have uniquely different requirements regarding state management in the finite state automata defined in refloat_thd.

I don't really see this as a big problem. You just output the `start_time`, `duration` and `type` of tone to play from `haptic_feedback_update()`. Then the application of the tone will be done in two separate places and slightly differently, but you got all the data you need.

For the alerts that are continuous in nature, like the pushbacks, you can keep the `duration` at say 300ms (which will give you a convenient minimum tone duration, though one that is applied even if the tone is long enough, but you can tweak that if you want) and continuously keep updating the `start_time` to keep the tone playing.

Relys: mmkay, so I factored that stuff out into a new function and also put placeholder comments where I think 6.05 stuff makes sense to be implemented (either updating pid or calling foc_play_tone).

I also moved set_current from main into utils since it needs to be able to be called from haptic_feedback. LMK if you have a better idea.

Relys:
Closing in favour of 6.05 implementation https://github.com/lukash/refloat/pull/12

PR #12, opened 2024-08-27
Feature: Implement Haptic Feedback 6.05

Relys:
All function calls should be < 6.05 safe as they explicitly check if the play tone pointers aren't null to determine if the API is available.

Features:
Duty Cycle Solid Tone: Duty Cycle threshold where solid tone is played. Tiltback will pulse at a period of 200ms and then go solid once this value is set. Set <= tiltback to always have it solid. Set >=95 to disable (just like tiltback)
Duty Cycle Strength: Zero disables. Will pulse at 200ms when duty cycle tiltback is present.
Duty Cycle Frequency:
Error Strength: Zero Disables. Will pulse at 200ms with a 400ms pause every 2 beeps for hv/lv. For temp (and bms in future) it will pulse at half rate. i.e. 100ms with 200ms pause.
Error Frequency:
Vibrate Strength: Zero disables. Plays in separate channel
Vibrate Frequency:

### DIFF
--- a/src/conf/datatypes.h
+++ b/src/conf/datatypes.h
@@ -185,6 +185,13 @@ typedef struct {
     float tiltback_lv;
     float tiltback_return_speed;
     float tiltback_constant;
+    float haptic_duty_solid;
+    float haptic_duty_volt;
+    int haptic_duty_freq;
+    float haptic_error_volt;
+    int haptic_error_freq;
+    float haptic_vibrate_volt;
+    int haptic_vibrate_freq;
     uint16_t tiltback_constant_erpm;
     float tiltback_variable;
     float tiltback_variable_max;
--- a/src/haptic_feedback.c
+++ b/src/haptic_feedback.c
@@ -1,0 +1,183 @@
+// Copyright 2024 Syler Clayton
+//
+// This file is part of the Refloat VESC package.
+//
+// Refloat VESC package is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by the
+// Free Software Foundation, either version 3 of the License, or (at your
+// option) any later version.
+//
+// Refloat VESC package is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+// or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program. If not, see <http://www.gnu.org/licenses/>.
+
+#include "haptic_feedback.h"
+
+void haptic_feedback_configure(HapticFeedback *haptic_feedback, RefloatConfig *float_conf) {
+    haptic_feedback->duty_solid = float_conf->haptic_duty_solid;
+    haptic_feedback->duty.volt = float_conf->haptic_duty_volt;
+    haptic_feedback->duty.freq = float_conf->haptic_duty_freq;
+    haptic_feedback->error.volt = float_conf->haptic_error_volt;
+    haptic_feedback->error.freq = float_conf->haptic_error_freq;
+    haptic_feedback->vibrate_volt = float_conf->haptic_vibrate_volt;
+    haptic_feedback->vibrate_freq = float_conf->haptic_vibrate_freq;
+}
+
+void haptic_feedback_reset(HapticFeedback *haptic_feedback, float current_time) {
+    if (haptic_feedback->tone_in_progress && VESC_IF->foc_stop_audio) {
+        VESC_IF->foc_stop_audio(true);
+    }
+    haptic_feedback->tone_in_progress = false;
+    haptic_feedback->timer = current_time;
+    haptic_feedback->alt_timer = current_time;
+    haptic_feedback->haptic_feedback_state.type = HAPTIC_FEEDBACK_NONE;
+    haptic_feedback->counter = 0;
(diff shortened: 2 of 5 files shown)


src/conf/datatypes.h
```diff
@@ -185,6 +185,13 @@ typedef struct {
     float tiltback_lv;
     float tiltback_return_speed;
     float tiltback_constant;
+    float haptic_duty_solid;
+    float haptic_duty_volt;
+    int haptic_duty_freq;
+    float haptic_error_volt;
+    int haptic_error_freq;
+    float haptic_vibrate_volt;
+    int haptic_vibrate_freq;
```
lukash, line 194: Can you make a struct member `haptic_feedback` the same way e.g. the LEDs have it's own config structs? Then you can easily make a copy of it in your `HapticFeedback` struct.

src/haptic_feedback.c
```diff
@@ -0,0 +1,183 @@
...
+    haptic_feedback->error.peroid = 200;  // 200 ms
+    haptic_feedback->note_duration = 0;
+    haptic_feedback->beep_num = 0;
+}
+
+void haptic_feedback_play_single_tone(float freq, float volt, int channel) {
+    if (!VESC_IF->foc_play_tone) {
+        return;
+    }
+    if ((volt > 0) && (freq > 0.0)) {
```
lukash, line 49: The braces around the `>` expressions are redundant (and this is a very common expression composition). Can you remove them? This repeats multiple times throughout the PR.

When in doubt, refer to [C Operator Precedence](https://en.cppreference.com/w/c/language/operator_precedence.)

src/haptic_feedback.c
```diff
@@ -0,0 +1,183 @@
...
+            haptic_feedback->haptic_feedback_state = haptic_feedback->error;
+            switch (state->sat) {
+            case SAT_PB_TEMPERATURE:
+                // Careful
+                if (haptic_feedback->haptic_feedback_state.peroid > 0) {
+                    haptic_feedback->haptic_feedback_state.peroid /= 2;
+                }
+                __attribute__((fallthrough));
+            case SAT_PB_LOW_VOLTAGE:
+                __attribute__((fallthrough));
```
lukash, line 94: The attribute isn't needed if the `case` labels follow directly after each other.

src/haptic_feedback.c
```diff
@@ -0,0 +1,183 @@
...
+                break;
+            }
+        }
+        // Reset state if a different type of event is coming in
+        if (haptic_feedback->haptic_feedback_state.type != prev) {
+            haptic_feedback_reset(haptic_feedback, current_time);
+        }
+
+        if (!haptic_feedback->tone_in_progress) {
+            haptic_feedback_play_two_tones(haptic_feedback);
```
lukash, line 112: In a clean design this call shouldn't be here. This Function should just set up the tones or the pattern, and the `haptic_feedback_update()` function take care of the playing by itself. I understand the logic you have relies on this, see my comment on the `haptic_feedback_update()` function.

src/haptic_feedback.c
```diff
@@ -0,0 +1,183 @@
...
+                    );
+                    haptic_feedback->beep_num--;
+                } else {
+                    haptic_feedback_play_two_tones(haptic_feedback);
+                }
+            } else {
+                VESC_IF->foc_stop_audio(false);
+            }
+        }
+    }
```
lukash, line 152: This logic is very complex and opaque. You also repeatedly call the play functions all the time, which is not very clean. What I'd do instead is take a similar approach to when you implement an animation. I would store a `bool` that says whether you are currently playing a tone. Then, depending on the time, determine whether you're supposed to be playing a tone, and if that differs from the current state, start or stop playing the tone.

For a duty tone pattern, tone-gap-tone-gap-..., you can do e.g.:
```C
const float tone_length = 0.2f;
const float period = tone_length  * 2; // note: according to convention I call period the whole segment that repeats
time = fmodf(time, period); // now time is your time within a single period
bool should_be_playing = time < tone_length;
```

For lets say three beeps and a pause:
```C
const float tone_length = 0.2f;
const float period = tone_length  * 8; // tone-gap-tone-gap-tone-gap-gap-gap = 8 beats
time = fmodf(time, period);
int beat = floorf(time / tone_length);
bool should_be_playing = (beat % 2) == 0 && beat != 6;
```

I hope this makes sense to you, if not let me know and I'll try to come up with a PoC to demonstrate better.

src/haptic_feedback.c
Relys: If you don't immediately start playing a tone when called, then by the time you get to the haptic_feedback_update function timer and alt_timer will already be out of date.

src/haptic_feedback.c
Relys: So I don't call any of the functions redundantly, only every period when alt_timer expires. Regarding the example for duty tone pattern, I'm not sure that this logic would blend smoothly with how I'm implementing creating a solid tone for when we've reached the duty cycle threshold (by continue to update alt_timer so it never reaches its next period). If you can provide an alternative that's fine, but I'm not sure I want to spend more cycles here rewriting this.

Jared-Is-Coding:
I would suggest these revisions. GitHub isn't letting me "review", so I'll just post the commit:

https://github.com/Jared-Is-Coding/refloat/commit/be7611a893b6ca8d1ce312d3bc3f813f1e88fb64

Will give this some actual ride testing this week.

PR #14, opened 2024-09-06
Reverse-Stop: smoother stop behavior (wife approved!)

surfdado:
Tail is now gently placed down in all situations
Timeouts extended to accomodate slow stopping
 - timeouts ensure we don't get stuck in limbo in corner cases (unlike other boards out there)

Also replaced floating point divisions with equivalent multiplications

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -315,7 +315,7 @@ static void configure(data *d) {
     d->max_duty_with_margin = VESC_IF->get_cfg_float(CFG_PARAM_l_max_duty) - 0.1;
 
     // Feature: Reverse Stop
-    d->reverse_tolerance = 50000;
+    d->reverse_tolerance = 20000;
     d->reverse_stop_step_size = 100.0 / d->float_conf.hertz;
 
     // Feature: Turntilt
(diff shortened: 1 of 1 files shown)


lukash:
LGTM, didn't test myself, I'll trust you on this.

Just once again a reminder to add the `Feature: ...` git trailer. I found out the trailer only works when there is no empty newline between your `Signed-off-by:` trailer and the other one (FTR, the `Signed-off-by:` doesn't really have a meaning in this repo, but obviously keep using it if you want).

Since you've got maybe more text for the trailer that can fit on one line, maybe refer to https://github.com/lukash/refloat/blob/main/CONTRIBUTING.md for how to do a multiline record. I know it's not nice and quite hairy, but it does the job in a way... You can run `./changelog.py` after you add the trailer to make sure it works and the markdown is reasoable (slight hand-tuning of the changelog is expected, it's not gonna be perfect the way the script spits it out).

Jared-Is-Coding:
Gave this some testing on a variety of angled surfaces - I found them all to be pretty a similar feel! This is also a significant improvement over the way the board currently sets down with reverse stop (a shaky mess).

Old reverse stop:
https://github.com/user-attachments/assets/5c40e7a8-f8ed-42cf-b129-4fcab48f241e

New reverse stop:
https://github.com/user-attachments/assets/99b3884e-ac92-4000-9db3-a0e9827214b8

Looks good.

surfdado:
Added the Feature: ... annotation

I will probably forget again next time I submit a PR...
But I added a reminder for adding "Feature:`..." in my "make clang" target I usually run before submitting a commit. Maybe that will help...

lukash:
Thanks! :call_me_hand:

lukash:
Shit, I totally missed you posted this against the 1.0 release... But why...
I'm gonna drop this from `release/v1.0` and push it to main.

surfdado:
> Shit, I totally missed you posted this against the 1.0 release... But why... I'm gonna drop this from `release/v1.0` and push it to main.

sorry, there was only two choices - main seemed experimental, so I chose 1.0
maybe create a release candidate branch?

lukash:
> sorry, there was only two choices - main seemed experimental, so I chose 1.0 maybe create a release candidate branch?

`main` is the development branch all PRs should be sent against. It is what 1.1 will become at this point, and the commit needs to get there in most cases (unless you'd be hotfixing something on a release maintenance branch, i.e. `release/v1.0`, which is not relevant on `main`). Then, in case you'd like the fix to also be backported (cherry-picked) to a release branch, you can ask for that and we can do that if it makes sense. Technically it can then be done either by me manually pushing the commit to the release branch or by you sending a second PR to that branch, but `main` should normally come first.

For this particular case, I think it's right on the line whether to backport to 1.0 or not, but I'm inclined to say no, just because I think the fix is somewhat complex and a bit more testing could be useful.

PR #16, opened 2024-09-13
Updates to CONTRIBUTING.md

Jared-Is-Coding:
Update code formatting (LLVM, lefthook, clang-format) installation
instructions to provide more details.

Move "Naming Conventions" information above formatting information, as
it contains the first steps for adding to the codebase

Make corrections to typos

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 Bugfix Pull Requests are always welcome. If you want to implement a feature, it is advised get in touch first, to make sure your feature is in line with the direction Refloat development is heading.
 
 ## Commit Messages
-Use descriptive (but succint) commit title. For non-trivial commits, add a short commit description with any due explanation.
+Use descriptive (but succinct) commit title. For non-trivial commits, add a short commit description with any due explanation.
 
 If the commit fixes a bug or adds a feature that should be mentioned in the changelog, add a `Fix:` or `Feature:` [git trailer](https://alchemists.io/articles/git_trailers) respectively.
 
@@ -60,8 +60,14 @@ CI runs on every push to main and needs to pass on every PR. It:
 - Runs a build of the package.
 - Checks formatting of C sources using `clang-format`.
 
+## Naming Conventions
+- Type names are in `PascalCase`.
+- Global constants and enum items are in `SCREAMING_CASE`.
+- Variables (local, global, struct members) and functions are in `snake_case`.
+- File names are in `snake_case.c`.
+
 ## Code Formatting
-C code is automatically formatted using `clang-format` version 18 (you need to provide this dependency yourself).
+C code is automatically formatted using `clang-format` version 18. You will need to provide this dependency yourself. See: [LLVM releases](https://releases.llvm.org/)
 
 The rest of the code (namely, the QML AppUI sources) is not auto-formatted. Please do your best to adhere to formatting you see surrounding the code you're editing.
 
@@ -73,7 +79,9 @@ clang-format -i src/*.{h,c}
 But, it's better to use git hooks.
 
 ## Git Hooks
-Refloat uses `lefthook` for managing git hooks. Again, you need to install this dependency in your development environment. Then run this to install the commit hook in the repo:
+Refloat uses `lefthook` for managing git hooks. Again, you will need to provide this dependency yourself. See: [`lefthook` on GitHub](https://github.com/evilmartians/lefthook)
+
+Once `lefthook` is installed, you can install the commit hook in the repo:
 ```sh
 lefthook install
 ```
@@ -84,9 +92,3 @@ To auto-fix all formatting of all files in your working tree, run:
 ```sh
 lefthook run clang-format-fix
 ```
-
-## Naming Conventions
-- Type names are in `PascalCase`.
-- Global constants and enum items are in `SCREAMING_CASE`.
-- Variables (local, global, struct members) and functions are in `snake_case`.
-- File names are in `snake_case.c`.


CONTRIBUTING.md
```diff
@@ -55,13 +55,19 @@ The `changelog.py` script is used to generate the final changelog (it's not perf
 - Use _rebase workflow_ for all PRs.
 - When addressing review comments, if possible, update original commits with fixes belonging to them. Then force-push.
 
+## Naming Conventions
+- Type names are in `PascalCase`.
+- Global constants and enum items are in `SCREAMING_CASE`.
+- Variables (local, global, struct members) and functions are in `snake_case`.
+- File names are in `snake_case.c`.
+
```
lukash, line 68: I liked the place naming conventions were before. Here you've put them inbetween Pull Requests and CI, I'm not sure how that fits logically...

CONTRIBUTING.md
```diff
@@ -55,13 +55,19 @@ The `changelog.py` script is used to generate the final changelog (it's not perf
...
 ## CI
 CI runs on every push to main and needs to pass on every PR. It:
 - Runs a build of the package.
 - Checks formatting of C sources using `clang-format`.
 
 ## Code Formatting
-C code is automatically formatted using `clang-format` version 18 (you need to provide this dependency yourself).
+C code is automatically formatted using `clang-format` version 18. You will need to provide this dependency yourself. See: [LLVM builds page](https://llvm.org/builds/) and [LLVM clang docs for integrations](https://releases.llvm.org/18.1.6/tools/clang/docs/ClangFormat.html)
```
lukash: I'm not sure how helpful those links are. I would expect most people to install `clang-format` primarily from their distro and I don't want to encourage downloading precompiled binaries, as IMO they should be the last resort thing, not first (I do wish there was a standard and automated way to specify and pull those dependencies, but there is not...).

And then for the integrations link, one thing is it's tied to a version so it adds to maintenance burden for when the version is updated. But more importantly I find it of little relevance to bootstrapping the build environment for the package itself. It contains various integration guidelines which are not really relevant to the package itself, you could add links to dozens of other useful setup articles if you wanted to go that way, but I don't think it belongs here.

CONTRIBUTING.md
```diff
@@ -73,7 +79,9 @@ clang-format -i src/*.{h,c}
 But, it's better to use git hooks.
 
 ## Git Hooks
-Refloat uses `lefthook` for managing git hooks. Again, you need to install this dependency in your development environment. Then run this to install the commit hook in the repo:
+Refloat uses `lefthook` for managing git hooks. Again, you will need to provide this dependency yourself. See: [`lefthook` on NPM](https://www.npmjs.com/package/lefthook)
```
lukash: Why the link to npm? It's a hipster js thing, not everyone is a hipster ok? :grin: On a more serious note, maybe a link to the github page would be better, it lists the various installation methods as well. I install lefhook from my distro.

Jared-Is-Coding: Updated with a link to the GitHub.

CONTRIBUTING.md
Jared-Is-Coding: Removed any references to a version, and instead linked to the base releases page (Which itself includes links to binaries, builds, and docs)

My thinking here is this: If someone is the type of person who wants to build their own packages, they will seek out the source for doing so. However, if someone is just interested in diving into development here, having links like this lowers the bar to entry just a little bit and makes it more inviting. That seems like a good thing to move toward.

If the concern is with encouraging download of precompiled packages, perhaps I can add a note about that as a "best practice", but I think many coming to this page would be aware already.

CONTRIBUTING.md
Jared-Is-Coding: Moved it one "spot" lower, before code formatting. I think it logically fits there, but I suppose it could stay down at the bottom as well. Either way is fine with me, would you prefer it at the bottom?

lukash: Honestly I prefer it where it was, because the Code Formatting section below is about the `clang-format` setup, which is part of setting the environment. The naming conventions is when you go edit the code by hand, and that to me fits better after the whole environment setup part. So I'd keep it at the end.

CONTRIBUTING.md
lukash: Well, I don't like just linking to the overall download page, because that makes it look like downloading the binary is the recommended/expected way to do it, while to me it's the very last resort. I know some people don't care too much, but it's technically a bad practice.

Also, these are contributing guidelines, it's expected you know your way around if you want to contribute code. And if not, you're willing to research it and figure it out.

Anyway, if you really want to add to this section (while, at the same time, not making it too long and bloat the contribution guidelines), please:
- List the packages to install on the common distros (Ubuntu and Fedora I guess). This is a bit complicated because sometimes the package is called `clang-tools` or `clang-tools-extra` (so this can be helpful information, but requires verifying the package names). I think the latter is the name for Fedora and I'm not entirely sure what's the situation on Ubuntu atm., I found booth the clang-tools and clang-format packages via a quick search. They also may or may not have a -18 in their name.
- Mention clang-format can be pulled as a dev environment dependency via other means.
- If nothing else is available, download the binary.

Jared-Is-Coding: > Also, these are contributing guidelines, it's expected you know your way around if you want to contribute code. And if not, you're willing to research it and figure it out.

Just trying to make it more approachable. I can see this isn't actually helpful, though. I think I'll just close this for the time being.

lukash: Some of the changes in this PR were helpful improvements...

PR #17, opened 2024-09-14
Use smooth duty universally

lukash:
Duty cycle can spike a lot, the purposes it's used for can all benefit from the mild smoothing we already have. Unify the usage and only use smooth duty for everything.

@surfdado putting this into a PR so that you have a chance to comment, WDYT? The motivation is smoother duty for the haptic duty alert (I had some spikes trigger it during normal riding). Please add a review if you don't mind.

Untested at the time of posting the PR, I'll only merge after I test it, but it's basically just the duty pushback and the wheelslip detection that uses the duty value.

### DIFF
--- a/src/motor_data.c
+++ b/src/motor_data.c
@@ -24,8 +24,6 @@
 #include <math.h>
 
 void motor_data_reset(MotorData *m) {
-    m->duty_smooth = 0;
-
     m->acceleration = 0;
     m->accel_idx = 0;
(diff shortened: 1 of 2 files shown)


surfdado:
I have no objections. The added delay is minimal (12ms), we could even be a bit more aggressive with the smoothing, like 0.95/0.05, without any downsides I can think of

PR #18, opened 2024-09-17
Haptic Feedback

lukash:
An alternative PR to #12. I believe this implementation is better structured and simpler.

The PR is not meant to be merged as-is, it's posted to allow for submitting feedback. It will be made available in a release as a feature preview and after gathering feedback on the feature we should reduce the number of config options.

### DIFF
--- a/src/haptic_feedback.c
+++ b/src/haptic_feedback.c
@@ -1,0 +1,131 @@
+// Copyright 2024 Lukas Hrazky
+//
+// This file is part of the Refloat VESC package.
+//
+// Refloat VESC package is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by the
+// Free Software Foundation, either version 3 of the License, or (at your
+// option) any later version.
+//
+// Refloat VESC package is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+// or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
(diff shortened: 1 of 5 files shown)


src/haptic_feedback.c
```diff
@@ -0,0 +1,127 @@
...
+    return 0;
+}
+
+void haptic_feedback_update(
+    HapticFeedback *hf, const State *state, float duty_cycle, float current_time
+) {
+    if (!VESC_IF->foc_play_tone) {
+        return;
+    }
+
```
Relys, line 93: You're gonna want to check if state->mode == MODE_FLYWHEEL so duty cycle haptics don't happen in flywheel mode.

src/haptic_feedback.c
```diff
@@ -0,0 +1,127 @@
...
+    hf->start_time = 0.0f;
+    hf->is_playing = false;
+}
+
+void haptic_feedback_configure(HapticFeedback *hf, const CfgHapticFeedback *cfg) {
+    hf->cfg = cfg;
+}
+
+static HapticFeedbackType state_to_haptic_type(const State *state) {
+    if (state->state != STATE_RUNNING) {
```
Relys, line 35: You're gonna want to check if state->mode == MODE_FLYWHEEL so duty cycle haptics don't happen in flywheel mode.

lukash: Duty SAT is not being set during Flywheel (and I tested it, it doesn't buzz for me). Not sure about the error ones right now. If they were:
- you'd supposedly want to hear them
- they'll most likely make the board fall over as they change the setpoint

Relys: Sounds good. Only reason I mention is because it was in the original float package.

PR #20, opened 2024-10-17
Flywheel safety: auto exit flywheel mode after 10s of being idle

surfdado:
Too many users have managed to ruin their package config by using flywheel mode and then not exiting the mode properly. This change uses the same strategy as we use for dark ride. Once it disengages the user has 10 seconds to put the board back into the upright/flywheel position, otherwise we return back to normal mode.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1443,6 +1443,10 @@ static void refloat_thd(void *arg) {
                 }
                 d->enable_upside_down = false;
                 d->state.darkride = false;
+                if (d->state.mode == MODE_FLYWHEEL) {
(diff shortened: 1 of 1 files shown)


lukash:
Just to comment here I brought up on discord that the 10s timeout is not very user friendly and I asked if anyone really re-engages the flywheel this way. I always turn it off, balance the board standing again and re-activate, the direct re-engage has never worked for me. So I was asking if we should just turn the flywheel off right away. Nico seemed to agree. Didn't get a response to that on discord @surfdado.

PR #23, opened 2024-10-21
Fix battery level param in COMMAND_GET_ALLDATA

Relys:
Fix for battery level parameter scaling so we actually get correct data. mc_get_battery_level is between 0.0 and 1.0 so we need to multiple by 100 first else we don't get the correct data.


lukash:
Thanks, could you just amend the commit title the way you amended the PR title and leave out the period at the end?

>  Fix battery level param in COMMAND_GET_ALLDATA

Relys:
Updated the commit message.

PR #24, opened 2024-11-01
LEDs: Add support for the WRGB/RGBW color order (WS2814 LED chip)

Jared-Is-Coding:
This adds LED Color Order options with support for the WRGB/RGBW order. Updated docstrings accordingly.

Feature: Add new LED Color Order configs
- Add LED Color Order options with support for the WRGB order (WS2814 LED chip).


src/conf/settings.xml
```diff
@@ -3160,6 +3160,8 @@ p, li { white-space: pre-wrap; }
             <valInt>0</valInt>
             <enumNames>GRB</enumNames>
             <enumNames>RGB</enumNames>
+            <enumNames>WRGB</enumNames>
+            <enumNames>RGBW</enumNames>
```
lukash: This now duplicates the RGB vs. RGBW values together with the two values in the LED types. Could you instead change the list to the following:
- GRB(W)
- RGB(W)
- (W)RGB

And keep the white presence in the logic based on the LED type value? I know it makes the logic more complicated, but it keeps the options not being redundant, your change makes them redundant and that'd be confusing for users.

lukash:
Also, this in the changelog record in the commit message:
> Allows for simplified bit OR logic.

Is not something relevant for the changelog and for users to read. Can you keep the changelog entry terse and just describe it from the user PoV, meaning, just say: "Added support for the (W)RGB color order (WS2814 LED chip).

Oh, and can you please also update the docstrings for the Color Order and LED Type options (the latter also mentions the supported LED Chip types, so let's update it as well).

Jared-Is-Coding:
@lukash I've implemented the requested methodology: White byte logic will continue to be separately managed.

Docstrings are now updated with related mentions of the WS2814 chip support.

lukash:
Thank you, code looks good now! Could you just make the changelog entry even shorter, it's a bit redundant, I think just a single line like this would suffice:
> Feature: Add the (W)RGB LED Color Order option (WS2814 LED chip).

Jared-Is-Coding:
> Feature: Add the (W)RGB LED Color Order option (WS2814 LED chip).

You're referring to the multiline Feature: tag right? Done.

lukash:
Yes, thank you!

PR #29, opened 2024-12-16
Simple-start improvements

surfdado:
Simple-start now includes a 1 second grace period before heel lifts get detected. Also the cool-down period after dismounting has been reduced to 2 seconds.

Feature: More tolerant simple-start implementation with 1 second grace period

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -148,7 +148,7 @@ typedef struct {
     float turntilt_target, turntilt_interpolated;
     float current_time;
     float dt;
-    float disengage_timer, nag_timer;
+    float engage_timer, disengage_timer, nag_timer;
     float idle_voltage;
     float fault_angle_pitch_timer, fault_angle_roll_timer, fault_switch_timer,
         fault_switch_half_timer;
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -409,6 +409,9 @@ static void reset_runtime_vars(data *d) {
     d->last_yaw_angle = 0;
     d->yaw_aggregate = 0;
 
+    // Reset engage_timer (keeps track of how long the board has been engaged/riding)
```
lukash, line 412: Subjective nitpick: I find this comment to be a bit too much of a:
```c
// add two ones together
int x = 1 + 1;
```

The code is self-explanatory IMO and the comment just mostly adds unnecessary verbosity. I can certainly merge this, just that I tend to remove this kind of comments while refactoring (and just try to keep the important ones).

lukash (changes requested):
The git trailer in your commit message doesn't work like this, git requires no newlines between trailers (I got bitten by this before, and with the amount of stuff piling up it's gonna be hard noticing a changelog line is missing):
```
Feature: More tolerant simple-start implementation with 1 second grace period
Signed-off-by: Dado Mista <dadomista@gmail.com>
```

Why do you use the Signed-off-by: one? It actually really only has meaning for the Linux kernel code and maybe few other repos where the owners assign a meaning to it. I mean maybe we could do that, I'm not sure it's necessary, I'd need to read up more on the legal stuff...

But most projects don't use it AFAIK :shrug:

lukash (approved):

PR #30, opened 2024-12-16
Improvements to HV behavior

surfdado:
In High Voltage cases:
1) make sure haptic buzz starts immediately (1st commit)
2) haptic buzz intensifies by up to 50% from 3s to 5s and remains at 150% level after 5s
3) tiltback kicks in with a 5s delay allowing for haptic buzz to convince the user to address the issue


src/main.c
```diff
@@ -763,12 +763,12 @@ static void calculate_setpoint_target(data *d) {
...
             }
-
-            d->state.sat = SAT_PB_HIGH_VOLTAGE;
         } else {
             // The rider has 500ms to react to the triple-beep, or maybe it was just a short spike
-            d->state.sat = SAT_NONE;
+            d->setpoint_target = 0;
         }
+        // setting the state regardless to ensure haptic buzz starts right away
+        d->state.sat = SAT_PB_HIGH_VOLTAGE;
```
lukash, line 771: Not super major, but for awareness: This function is bound for a refactor, I'd like to change it so that only the `state.sat` is set based on the conditions, and then there'll be a function that sets `setpoint_target` according to `sat` (haven't fully thought through some of the cases but that's the rough idea).

This special-case throws a wrench into that plan, and is not pretty as it is. Obviously if this behavior is desired it can be worked out and it should be prettier after the refactor... But it's better to not have special cases.

src/haptic_feedback.c
```diff
@@ -145,9 +148,16 @@ void haptic_feedback_update(
...
+            float strength = scale_strength(tone->strength, hf->cfg, speed);
+            if (type_to_play == HAPTIC_FEEDBACK_ERROR_HI_VOLTAGE) {
+                float duration = current_time - hf->start_time;
+                if (duration > 3) {
+                    // more than 3 seconds in HV? Start scaling up for next 2 seconds
+                    // up to 50% louder
+                    strength = strength * fminf(1.5, duration * 0.25 + 0.25);
+                }
+            }
+            VESC_IF->foc_play_tone(0, tone->frequency, strength);
```
lukash, line 160: This is a second special-case being added :smiling_face_with_tear:.

Also, this raises the strength above the user-set threshold, and by as much as 50%. Doesn't seem like a good thing to do.

If I was to contrive a disaster scenario, someone configures their haptic feedback and tests it at the configured levels (no way to easily know what this +50% strength will do and even that it can actually happen on HV). Then, months later, they trigger HV on a steep downhill. They don't react in time, the loud haptics mess up the imu and make the nose go higher, they go into taildrag and crash.

src/haptic_feedback.h
```diff
@@ -27,7 +27,8 @@ typedef enum {
     HAPTIC_FEEDBACK_DUTY,
     HAPTIC_FEEDBACK_DUTY_CONTINUOUS,
     HAPTIC_FEEDBACK_ERROR_TEMPERATURE,
-    HAPTIC_FEEDBACK_ERROR_VOLTAGE,
+    HAPTIC_FEEDBACK_ERROR_LO_VOLTAGE,
+    HAPTIC_FEEDBACK_ERROR_HI_VOLTAGE,
```
lukash, line 31: I don't think the one or two extra chars will change anything, so I'd prefer the `HIGH` and `LOW` spelling.

src/main.c
```diff
@@ -755,9 +755,9 @@ static void calculate_setpoint_target(data *d) {
     } else if (d->motor.duty_cycle > 0.05 && input_voltage > d->float_conf.tiltback_hv) {
         d->beep_reason = BEEP_HV;
         beep_alert(d, 3, false);
-        if (((d->current_time - d->tb_highvoltage_timer) > .5) ||
-            (input_voltage > d->float_conf.tiltback_hv + 1)) {
-            // 500ms have passed or voltage is another volt higher, time for some tiltback
+        if ((d->current_time - d->tb_highvoltage_timer) > 5) {
+            // It is assumed that haptic feedback is enabled!
+            // 5s have passed or voltage is another volt higher, time for some tiltback
```
lukash, line 760: You've removed the "another volt higher" part of the condition. A case in point with these comments, self-explanatory code is better.

But more importantly, at this stage we can't make that assumption, which in my eyes just invalidates the PR.

lukash (changes requested):
This PR looks problematic to me, let's discuss on discord.

PR #32, opened 2024-12-22
Parking mode idle improved behavior

surfdado:
If rider jumps off the board and the board changes direction after jumping off, it is assumed that the parking brake should be on.

This addresses the use case where you're going uphill and you jump off because it got too steep

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -559,6 +559,8 @@ static bool check_faults(data *d) {
             // allow turning it off by engaging foot sensors
             state_stop(&d->state, STOP_SWITCH_HALF);
             return true;
+        } else {
+            d->motor.last_erpm_sign = d->motor.erpm_sign;
         }
     } else {
         bool disable_switch_faults = d->float_conf.fault_moving_fault_disabled &&
@@ -592,6 +594,7 @@ static bool check_faults(data *d) {
             }
         } else {
             d->fault_switch_timer = d->current_time;
+            d->motor.last_erpm_sign = d->motor.erpm_sign;
         }
 
         // Feature: Reverse-Stop
@@ -1453,9 +1456,7 @@ static void refloat_thd(void *arg) {
             break;
         }
 
-        motor_control_apply(
-            &d->motor_control, d->motor.abs_erpm_smooth, d->state.state, d->current_time
-        );
+        motor_control_apply(&d->motor_control, &d->motor, d->state.state, d->current_time);
         VESC_IF->sleep_us(d->loop_time_us);
     }
 }
--- a/src/motor_control.c
+++ b/src/motor_control.c
@@ -38,6 +38,12 @@ void motor_control_configure(MotorControl *mc, const RefloatConfig *config) {
     mc->click_current = config->startup_click_current;
     mc->parking_brake_mode = config->parking_brake_mode;
     mc->main_freq = config->hertz / 2;
+
+    if (mc->parking_brake_mode == PARKING_BRAKE_ALWAYS) {
+        mc->parking_brake_active = true;
+    } else if (mc->parking_brake_mode == PARKING_BRAKE_NEVER) {
+        mc->parking_brake_active = false;
+    }
 }
 
 void motor_control_request_current(MotorControl *mc, float current) {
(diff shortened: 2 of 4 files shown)


src/main.c
```diff
@@ -592,6 +594,7 @@ static bool check_faults(data *d) {
             }
         } else {
             d->fault_switch_timer = d->current_time;
+            d->motor.last_erpm_sign = d->motor.erpm_sign;
```
lukash, line 597: Modifying the `MotorData` structure outside of the `motor_data_` functions goes against the concept of the modularization (which is, any work with the module data happens inside the module, outside of it, you only read them).

I'm also not sure there's more than meets the eye to the surgical precision with which you placed the assignment in two places of the `check_faults()` function :grinning:. But the fact it's two places is bad in itself, and this function really needs stuff removed, not added (it needs a refactor).

Instead, please set `last_erpm_sign` inside `motor_data_update()`. You'll have to pass whatever data you need for the condition. Seems to me you need `d->state.state` and check it's equal to `STATE_RUNNING` and keep updating the var for as long as that holds true. But I might have missed a detail in how you placed those assignments, it's not entirely obvious (which is the point I'm trying to make).

I'd call the variable `erpm_last_running_sign`.

src/motor_control.c
```diff
@@ -38,6 +38,13 @@ void motor_control_configure(MotorControl *mc, const RefloatConfig *config) {
     mc->click_current = config->startup_click_current;
     mc->parking_brake_mode = config->parking_brake_mode;
     mc->main_freq = config->hertz / 2;
+
+    // set parking brake true/false here instead of each cycle
```
lukash: The "... instead of each cycle" refers to the change you're making, not the semantics of the code. It's irrelevant outside the context of this PR, please remove the comment, it's only going to be confusing when encountered after the PR is merged.

src/motor_control.c
```diff
@@ -51,14 +58,7 @@ static inline void reset_tone(MotorControl *mc) {
...
 
-void motor_control_apply(MotorControl *mc, float abs_erpm, RunState state, float time) {
-    if (mc->parking_brake_mode == PARKING_BRAKE_ALWAYS ||
-        (mc->parking_brake_mode == PARKING_BRAKE_IDLE && state != STATE_RUNNING && abs_erpm < 50)) {
-        mc->parking_brake_active = true;
-    } else if (mc->parking_brake_mode == PARKING_BRAKE_NEVER || state == STATE_RUNNING) {
-        mc->parking_brake_active = false;
-    }
-
+void motor_control_apply(MotorControl *mc, MotorData *md, RunState state, float time) {
```
lukash: Can you make it `const MotorData *md`? That ensures the `MotorData` struct can't be modified in the function, it's very helpful to enforce this and anyone reading the code can have immediate certainty the struct is not being messed with here.

src/motor_control.c
```diff
@@ -87,7 +87,17 @@ void motor_control_apply(MotorControl *mc, float abs_erpm, RunState state, float
         VESC_IF->mc_set_current(mc->requested_current);
     } else {
         // Brake logic
-        if (abs_erpm > ERPM_MOVING_THRESHOLD) {
+        if (mc->parking_brake_mode == PARKING_BRAKE_IDLE) {
+            // only evaluate parking brake each cycle if pbmode is IDLE
+            if ((state != STATE_RUNNING) &&
+                ((md->abs_erpm_smooth < 50) || (md->last_erpm_sign != md->erpm_sign))) {
```
lukash: Please don't use brackets around expressions when they are "obviously" not necessary, they add visual clutter. What's obvious is obviously subjective, but these are very common cases and everyone should know the general precedence. If in doubt, just consult https://en.cppreference.com/w/c/language/operator_precedence.

The condition here should be:
```c
if (state != STATE_RUNNING && (md->abs_erpm_smooth < 50 || md->last_erpm_sign != md->erpm_sign)) {
```

src/motor_control.c
```diff
@@ -87,7 +87,17 @@ void motor_control_apply(MotorControl *mc, float abs_erpm, RunState state, float
...
+                mc->parking_brake_active = true;
+            } else {
+                mc->parking_brake_active = false;
+            }
+        }
```
lukash: I like this change to the logic (where the ALWAYS/NEVER cases are handled in configure), it's easier to understand this way :+1:.

But is there a reason you put this into the "brake logic" branch, meaning it's not evaluated when there's regular current requested? I'm not sure of the implications right now...

lukash (changes requested):
Thanks for this, I like it in general, comments are mostly about putting things in the right places and nitpicks.

lukash:
Forgot to mention, the changelog record:

> Fix: don't let board roll downhill after jumping off at low speed

Will be out of context and not really clear in the changelog. I'd also incline more towards marking it as a feature, as it's not a bug that it's fixing per se. What about something like:
> Feature: Parking Brake: Brake immediately if the board would roll the opposite way (e.g. downhill when one jumps off while going uphill)

src/motor_control.c
JJJ: @surfdado's approach is faster to understand at a glance, by encapsulating each comparison inside of its own _intentional_ set of brackets.

Having everything inline including one set of brackets while omitting brackets deeper than 2 suggests they were missed.

I'd completely go the other direction and prefer that each comparison & group be its own indentation.

White space costs nothing, and only makes code easier for human eyes to interpret. <3

lukash: Sorry, this is subjective, to me it makes it harder to interpret (because I need to carefully verify each bracket encapsulation is correct, sometimes even resort to counting them, whereas if they're not there, it's much clearer to me).

> White space costs nothing ...

Not true, space costs exactly that, space on the screen. Extra brackets make lines longer, may make some lines need to be wrapped, greatly reducing readability. Any redundant extra characters increase size while not increasing information, therefore reducing entropy.

(And this is as far as I want to go into this discussion :sweat_smile:)

surfdado:
Reposting the updated PR against the testing branch

lukash:
Closing in favor of the replacement PR: https://github.com/lukash/refloat/pull/34

PR #33, opened 2024-12-25
Negative variable tiltback via negative rate

lukash:
Changing negative Variable Tiltback to be configured via a negative rate, not negative maximum. Also a small refactor of some recalculations of config values.


lukash:
I chose to not change the behavior and use the Max with a sign, I'm just renaming it to "Target" and doing a few other improvements to the config options. Already in the `testing` branch, closing this.

PR #34, opened 2025-01-11
Parking brake checks for erpm sign now

surfdado:
If rider jumps off the board and the board changes direction after jumping off, it is assumed that the parking brake should be on.

Feature: Parking Brake: Brake immediately if the board would roll the opposite way (e.g. downhill when one jumps off while going uphill)


src/motor_control.c
```diff
@@ -51,12 +57,14 @@ static inline void reset_tone(MotorControl *mc) {
     mc->tone_high = 0;
 }
 
-void motor_control_apply(MotorControl *mc, float abs_erpm, RunState state, float time) {
-    if (mc->parking_brake_mode == PARKING_BRAKE_ALWAYS ||
-        (mc->parking_brake_mode == PARKING_BRAKE_IDLE && state != STATE_RUNNING && abs_erpm < 50)) {
-        mc->parking_brake_active = true;
-    } else if (mc->parking_brake_mode == PARKING_BRAKE_NEVER || state == STATE_RUNNING) {
-        mc->parking_brake_active = false;
+void motor_control_apply(MotorControl *mc, const MotorData *md, RunState state, float time) {
```
JJJ, line 60: `const MotorData *motor` is used elsewhere, though I personally think `*md` is best:

https://github.com/lukash/refloat/blob/fe9d7b4a2c80432d4913a2c58e58eaf2637c78b9/src/torque_tilt.c#L34

https://github.com/lukash/refloat/blob/fe9d7b4a2c80432d4913a2c58e58eaf2637c78b9/src/atr.c#L53

https://github.com/lukash/refloat/blob/fe9d7b4a2c80432d4913a2c58e58eaf2637c78b9/src/atr.c#L203

https://github.com/lukash/refloat/blob/fe9d7b4a2c80432d4913a2c58e58eaf2637c78b9/src/torque_tilt.h#L35

----

(`*m` used in other places, so maybe it isn't important, yet?)

https://github.com/lukash/refloat/blob/fe9d7b4a2c80432d4913a2c58e58eaf2637c78b9/src/motor_data.c#L26

https://github.com/lukash/refloat/blob/fe9d7b4a2c80432d4913a2c58e58eaf2637c78b9/src/motor_data.c#L39

src/motor_control.c
```diff
@@ -51,12 +57,14 @@ static inline void reset_tone(MotorControl *mc) {
...
+    if (mc->parking_brake_mode == PARKING_BRAKE_IDLE) {
+        if (state != STATE_RUNNING &&
+            (md->abs_erpm_smooth < 50 || md->last_erpm_sign != md->erpm_sign)) {
```
JJJ, line 63: Is `50` an approximation?

Should this be variable/configurable eventually?

src/motor_data.h
```diff
@@ -30,6 +30,7 @@ typedef struct {
     float abs_erpm_smooth;
     float last_erpm;
     int8_t erpm_sign;
+    int8_t last_erpm_sign;  // last erpm sign prior to footpads disengaging
```
JJJ, line 33: @lukash is probably going to tell you to delete your code comment 😂 because none of the others have them

surfdado: we need more comments, not less

src/motor_control.c
surfdado: 1) this isn't part of my PR
2) but PLEASE not another parameter for something as mundane as this. 50 seems just fine, if you have arguments why it should be different let's hear them

src/motor_control.c
surfdado: I tried to be consistent within the line of code I changed, so *md makes sense. But idk why you would even bother commenting on stuff like this. I could care less. Let Lukas name it however he wants.

src/motor_data.h
JJJ: I agree, but not everyone does

src/motor_control.c
JJJ: > idk why you would even bother commenting on stuff like this

Because that is how contributing to open-source software works, and how the code gets better over time.

The world reviews requests for progress, and iterates forward together.

>  I could care less

That's your problem; don't make it mine.

src/motor_control.c
JJJ: > if you have arguments why it should be different let's hear them

I'm not arguing, or suggesting – I'm asking, while reviewing/observing.

> this isn't part of my PR

You're touching this line-of-code and suggesting a change to the surrounding behavior. While we're all here now, look at it, consider it, understand whether `50` is a problem.

It stands out to me as a code-smell that may be contributing to the parking brake feature not behaving as intended, because `50` erpm is unlikely to just be the magic value that makes it work perfectly.

I'm not trying to block your progress or be a jerk, so chill with the defensiveness eh?

src/motor_control.c
lukash: It would be nice to have those consistent, but I'm not gonna spend time going over them to try to unify everything. Sometimes it's a balance of a short name that'll keep the lines from wrapping vs. the name being descriptive enough.

And I'm not gonna bug the PR author with such a small thing if the name is not outright bad.

If you want, @JJJ, you can post a PR unifying all the names of these arguments, I'll merge it :grin:

src/motor_control.c
lukash: I put the 50 there. Seemed to work well to me. This value will not be configurable, it's supposed to represent "0" with a small amount of leeway for erpm noise.

src/motor_data.h
lukash: This comment is good, because it's not clear what "last" means in the variable name. It's totally fine this one has a comment and the others don't.

If I was to be extra nitpicky, saying just _"... prior to disengaging"_ would be a bit better.

lukash:
@surfdado thanks for addressing the review, but the actual most important comment was not addressed: https://github.com/lukash/refloat/pull/32#discussion_r1897320143

We need well-structured code more than we need e.g. comments.

Also, is this working well for you? In situations when I had to jump off the board when it wouldn't go up a steep hill, most of the time I'd push the board on the tail just a bit before disengaging, making it shoot downhill. I don't think this implementation would catch it, as it would start the reverse before disengaging.

What I was imagining when we discussed it was `last_erpm_sign` with a delay, e.g. 1 second. Maybe even 0.5s would do. Not entirely sure about it, I'd have to test it, it'd obviously be a bit harder to implement and this PR is an improvement already, so just something for you to consider if you'd like.

surfdado:
yes, this feature works as it should now for me - it no longer shoots downhill for me on steep climbs when I need to dismount. But of course, if you're already sliding backwards by the time you jump off this feature won't work - and I don't really see a good way around that. A timer based implementation seems hacky to my but if you'd like to do it that way please go ahead.

lukash:
Yes, it would not be pretty and could be considered hacky. I just wanted to bring it up, I won't go testing or doing it now, maybe sometime later, after I see how your implementation does.

Meanwhile, my one remaining point blocking the PR remains without an answer... :disappointed:

PR #36, opened 2025-02-22
Set 0 current while state == DISABLED

surfdado:
Fix: when DISABLED release the brake and allow tool based motor control

### DIFF
--- a/src/motor_control.c
+++ b/src/motor_control.c
@@ -22,6 +22,7 @@
 #include "vesc_c_if.h"
 
 void motor_control_init(MotorControl *mc) {
+    mc->disabled = false;
     mc->current_requested = false;
     mc->requested_current = 0.0f;
     mc->click_counter = 0;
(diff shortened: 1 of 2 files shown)


lukash:
Why are you submitting this against the `v1.1-preview2` branch?

surfdado:
Well the preview is broken and needs fixing... but like I said, if you got your own fix just update the preview with it

lukash:
I was not planning on fixing this in the preview, I consider it a minor problem. I never re-run the motor detection and I don't really see the need for it myself.

Adding a commit on the branch by itself won't make the release happen and it makes zero sense applying different fixes to different branches, so your PR submission without a single word of explanation baffles me quite bit to be honest.

Feel free to explain yourself and actually make a case for the release and I can do it.

Did you try my fix yet?

lukash:
Ultimately adopted this approach for the `main` branch, closing...

PR #37, opened 2025-03-10
Feature: Add Support for BMS Tiltback 2.0

Relys:
Rewritten to use variables in Refloat configuration for all parameters as an alternative to https://github.com/lukash/refloat/pull/27 . Made lisp portion much smaller and am just going to do processing in C (need to anyways to access float_conf). Lisp takes up eeprom space and is not compiler optimized like C.

There are no hard coded parameters now. All comments were addressed from other PR with this rewrite. All features tested and confirmed working.

Requires 6.0+ firmware.

6.06 has easy methods that I merged upstream that are more efficient and will use those if detected:

```
(get-bms-val 'bms-v-cell-min)
(get-bms-val 'bms-v-cell-max)
```

Also on 6.06 it will also attempt to get the actual minimum cell temperature (as opposed to just using the max cell temperature as a fall back) as well as the bms mosfet temperature. This relies on the BMS being updated to the new protocol which can be checked with (get-bms-val 'bms-data-version). If you want to learn more please refer to LispBM manual.

Everything will provide tiltback except the cell balance. It will start beeping 10 second after disengage (giving it time to settle) before alerting if criteria is met.

### DIFF
--- a/package.lisp
+++ b/package.lisp
@@ -8,6 +8,13 @@
 ; Set firmware version:
 (apply ext-set-fw-version (sysinfo 'fw-ver))
 
+; Init bms
+(if (>= (first (sysinfo 'fw-ver)) 6) {
+    (import "src/bms.lisp" 'bms)
+    (read-eval-program bms)
+    (spawn "bms" 50 init-bms)
+})
+
 ; Set to 1 to monitor debug variables
 (define debug 1)
 
--- a/src/bms.h
+++ b/src/bms.h
@@ -1,0 +1,36 @@
+// Copyright 2024 Syler Clayton
+//
+// This file is part of the Refloat VESC package.
+//
+// Refloat VESC package is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by the
+// Free Software Foundation, either version 3 of the License, or (at your
+// option) any later version.
+//
+// Refloat VESC package is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+// or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+// more details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program. If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <stdbool.h>
+#include <stdint.h>
+
+typedef enum {
+    BMSF_NONE = 0,
+    BMSF_CONNECTION = 1,
+    BMSF_OVER_TEMP = 2,
+    BMSF_CELL_OVER_VOLTAGE = 3,
+    BMSF_CELL_UNDER_VOLTAGE = 4,
+    BMSF_CELL_OVER_TEMP = 5,
+    BMSF_CELL_UNDER_TEMP = 6,
+    BMSF_CELL_BALANCE = 7
+} BMSFaultCode;
+
+bool bms_get_fault(uint32_t fault_mask, BMSFaultCode fault_code);
+
+void bms_set_fault(uint32_t *fault_mask, BMSFaultCode fault_code);
--- a/src/bms.lisp
+++ b/src/bms.lisp
@@ -1,0 +1,39 @@
+
+(defun init-bms () {
+    (var v-cell-support (eq (first (trap (get-bms-val 'bms-v-cell-min))) 'exit-ok))
+    (var temp-max 0)
+    (var temp-min 0)
+    (var v-min 0)
+    (var v-max 0)
+    (var temp-fet -281)
+    (var num-cells 0)
+    (var cell-v 0.0)
+    (loopwhile t {
+        (if (and (>= (get-bms-val 'bms-can-id) 0) (ext-bms)) {
+            (setq temp-max (get-bms-val 'bms-temp-cell-max))
+            (setq temp-min temp-max)
+            
+            (if v-cell-support {
+                (if (= (get-bms-val 'bms-data-version) 1) {
+                    (setq temp-min (get-bms-val 'bms-temps-adc 1))
+                    (setq temp-fet (get-bms-val 'bms-temps-adc 3))
+                })
+                (setq v-min (get-bms-val 'bms-v-cell-min))
+                (setq v-max (get-bms-val 'bms-v-cell-max))
+            } {
+                (setq num-cells (get-bms-val 'bms-cell-num))
+                (if (> num-cells 0) {
+                    (setq v-min (get-bms-val 'bms-v-cell 0))
+                    (setq v-max v-min)
+                    (looprange i 1 num-cells {
+                        (setq cell-v (get-bms-val 'bms-v-cell i))
+                        (if (< cell-v v-min) (setq v-min cell-v))
+                        (if (> cell-v v-max) (setq v-max cell-v))
+                    })
+                })
+            })
+            (ext-bms v-min v-max temp-min temp-max temp-fet (get-bms-val 'bms-msg-age))
+        })
+        (sleep 0.2)
+    })
+})
--- a/src/conf/datatypes.h
+++ b/src/conf/datatypes.h
@@ -193,6 +193,14 @@ typedef struct {
     float tiltback_variable;
     float tiltback_variable_max;
     uint16_t tiltback_variable_erpm;
+    bool tiltback_bms_enabled;
+    float tiltback_cell_lv;
+    float tiltback_cell_hv;
+    int8_t tiltback_cell_lt;
+    int8_t tiltback_cell_ht;
+    int8_t tiltback_bms_ht;
+    uint8_t tiltback_bms_msg;
+    float tiltback_cell_bal;
     FLOAT_INPUTTILT_REMOTE_TYPE inputtilt_remote_type;
     float inputtilt_speed;
     float inputtilt_angle_limit;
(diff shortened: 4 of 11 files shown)


Relys:
Changed some of the defaults, added bms connection error beeps while idle and also moved that and balance beeps to 10 seconds instead of 1

package.lisp
```diff
@@ -8,6 +8,13 @@
 ; Set firmware version:
 (apply ext-set-fw-version (sysinfo 'fw-ver))
 
+; Init bms
+(if (>= (first (sysinfo 'fw-ver)) 6) {
```
lukash, line 12: Is this check even relevant? Firmwares before 6.0 didn't have package support, right?

package.lisp
```diff
@@ -8,6 +8,13 @@
...
+    (import "src/bms.lisp" 'bms)
```
lukash, line 13: If the code were to be in a separate file, certainly not in `/src`, likely `/lisp`, but the code is pretty sort, I'd vote for inlining it in `package.lisp` rather than creating a new directory.

package.lisp
```diff
@@ -8,6 +8,13 @@
...
+    (read-eval-program bms)
+    (spawn "bms" 50 init-bms)
```
lukash, line 15: Please change the name to "Refloat BMS" (there's already "Refloat Main" and "Refloat Aux" from C).

src/bms.h
```diff
@@ -0,0 +1,36 @@
...
+    BMSF_CONNECTION = 1,
+    BMSF_OVER_TEMP = 2,
+    BMSF_CELL_OVER_VOLTAGE = 3,
+    BMSF_CELL_UNDER_VOLTAGE = 4,
+    BMSF_CELL_OVER_TEMP = 5,
+    BMSF_CELL_UNDER_TEMP = 6,
+    BMSF_CELL_BALANCE = 7
+} BMSFaultCode;
+
+bool bms_get_fault(uint32_t fault_mask, BMSFaultCode fault_code);
```
lukash, line 34: I think conventionally this function would be better named `bms_is_fault`.

src/bms.lisp
```diff
@@ -0,0 +1,39 @@
+
+(defun init-bms () {
+    (var v-cell-support (eq (first (trap (get-bms-val 'bms-v-cell-min))) 'exit-ok))
+    (var temp-max 0)
+    (var temp-min 0)
+    (var v-min 0)
+    (var v-max 0)
```
lukash, line 7: Nitpick :grinning: could it be `v-max` before `v-min` just like `temp-max` before `temp-min`?

src/bms.lisp
```diff
@@ -0,0 +1,39 @@
...
+                    (setq v-min (get-bms-val 'bms-v-cell 0))
+                    (setq v-max v-min)
+                    (looprange i 1 num-cells {
+                        (setq cell-v (get-bms-val 'bms-v-cell i))
+                        (if (< cell-v v-min) (setq v-min cell-v))
+                        (if (> cell-v v-max) (setq v-max cell-v))
+                    })
+                })
+            })
+            (ext-bms v-min v-max temp-min temp-max temp-fet (get-bms-val 'bms-msg-age))
```
lukash, line 35: I think it'd be better to read the age at the very beginning. It's not that the code is very robust against races, but in case of a bad connection you could read zeros and then read a recent timestamp if an update just occurred. That would report as an assortment of temp/voltage alerts, instead of a connection issue...

src/conf/datatypes.h
```diff
@@ -193,6 +193,14 @@ typedef struct {
...
     float tiltback_variable_max;
     uint16_t tiltback_variable_erpm;
+    bool tiltback_bms_enabled;
+    float tiltback_cell_lv;
+    float tiltback_cell_hv;
+    int8_t tiltback_cell_lt;
+    int8_t tiltback_cell_ht;
+    int8_t tiltback_bms_ht;
+    uint8_t tiltback_bms_msg;
+    float tiltback_cell_bal;
```
lukash, line 203: These should go into a struct, I'm thinking like this:
```c
typedef struct {
    bool enabled;
    float cell_lv_threshold;
    float cell_hv_threshold;
    int8_t cell_ltemp_threshold;
    int8_t cell_htemp_threshold;
    int8_t bms_htemp_threshold;
    float cell_balance_diff;
} CfgBMS;

typedef struct {
    // ...
    CfgBMS bms;
    // ...
} RefloatConfig;
```

Note I removed `tiltback_bms_msg`, I think we should hardcode that.

And I don't want these in the pushback alerts section, I'd like to move towards calling these just Alerts and pushback be one of the ways for alerts to be performed, next to haptics.

I'm thinking best to put these in a separate tab called "BMS", there's enough options to warrant it.

src/main.c
```diff
@@ -226,6 +234,12 @@ static void configure(Data *d) {
         sizeof(headlights_off_konami_sequence)
     );
 
+    if (d->float_conf.tiltback_bms_enabled) {
+        d->bms_fault = BMSF_CONNECTION;
+    } else {
+        d->bms_fault = BMSF_NONE;
+    }
```
lukash, line 241: Why this? IMO it should be initialized to `NONE` and only set the connection error on a timeout.

src/main.c
```diff
@@ -612,6 +635,17 @@ static void calculate_setpoint_target(Data *d) {
...
+    } else if (bms_get_fault(d->bms_fault, BMSF_CONNECTION)) {
+        beep_alert(d, 3, true);
+        d->beep_reason = BEEP_BMS_CONNECTION;
+
+        if (d->motor.erpm > 0) {
+            d->setpoint_target = d->float_conf.tiltback_hv_angle;
+        } else {
+            d->setpoint_target = -d->float_conf.tiltback_hv_angle;
+        }
+        d->state.sat = SAT_PB_BMS_CONNECTION;
```
lukash, line 648: I'm not sure about adding another SAT at all, but if we do, I'd call it just generic `SAT_PB_ALERT` with the intention to use it for any sort of alert besides the ones already present. But not sure this needs a pushback, more so seeing you're also doing a beeper when in `STATE_READY` - a buzz would probably suffice, but the current alerting infra is very limiting.

src/main.c
```diff
@@ -927,6 +986,17 @@ static void refloat_thd(void *arg) {
...
+                // Alert user if cells are out of balance
+                if (bms_get_fault(d->bms_fault, BMSF_CELL_BALANCE)) {
+                    beep_alert(d, 1, false);
+                    d->beep_reason = BEEP_CELL_BALANCE;
+                }
+                // Alert user if bms connection failed.
+                if (bms_get_fault(d->bms_fault, BMSF_CONNECTION)) {
+                    beep_alert(d, 1, true);
+                    d->beep_reason = BEEP_BMS_CONNECTION;
+                }
```
lukash, line 999: I already commented on this on the previous PR, seems to me this will be beeping continuously, so first off, did you test it?

I'll need to think more about the alerting system, and how to get this in now in some sensible way and not have it wait for some sort of alerting revamp.

src/main.c
```diff
@@ -2042,6 +2112,43 @@ static lbm_value ext_set_fw_version(lbm_value *args, lbm_uint argn) {
     return VESC_IF->lbm_enc_sym_true;
 }
 
+// Called from Lisp to pass in values from the bms. If no prameters are called, return
+// tiltback_bms_enabled.
+static lbm_value ext_bms(lbm_value *args, lbm_uint argn) {
+    Data *d = (Data *) ARG;
+    d->bms_fault = BMSF_NONE;
```
lukash, line 2119: The way you update `d->bms_fault` by resetting it here and later setting the error flags again is not atomic and will lead to unnecessary races. `d->bms_fault` should be updated in a single operation to make it atomic on the stm32. Otherwise, if there is e.g. some timing hooked onto some of the faults, like a timeout on the HV pushback, it could get accidentally reset if a loop iteration runs between the above reset and the set of the fault that follows below.

src/main.c
```diff
@@ -2042,6 +2112,43 @@ static lbm_value ext_set_fw_version(lbm_value *args, lbm_uint argn) {
...
+    if (VESC_IF->lbm_dec_as_i32(args[3]) > d->float_conf.tiltback_cell_ht) {
+        bms_set_fault(&d->bms_fault, BMSF_CELL_OVER_TEMP);
+    }
+    if (VESC_IF->lbm_dec_as_i32(args[4]) > d->float_conf.tiltback_bms_ht) {
+        bms_set_fault(&d->bms_fault, BMSF_OVER_TEMP);
+    }
+    if (fabsf(VESC_IF->lbm_dec_as_float(args[0]) - VESC_IF->lbm_dec_as_float(args[1])) >
+        d->float_conf.tiltback_cell_bal) {
+        bms_set_fault(&d->bms_fault, BMSF_CELL_BALANCE);
+    }
```
lukash, line 2147: I understand putting the logic here is the most straightforward, but there should be a separation between the passing of the data from lisp and the evaluation logic. What seems the best approach to me is declaring a struct with all the values passed in from lisp. In this function, just fill the struct with values and then have a function like `bms_update_fault(uint32_t *fault_mask, TheNewStruct *bms_data);`

lukash:
I've addressed the points from the review and made some more adjustments to finish this: https://github.com/lukash/refloat/tree/bms-pushback

The commit from this PR was used with very minor (zero functional impact) adjustments and amended in subsequent commits.

It's ready for merging from code standpoint. I tested the BMS communication on 6.05 as well as as much logic as I could, but couldn't test the alerting while riding as my board with a CAN BMS isn't in a rideable state right now. After it gets some testing I'll proceed with merging and releasing it.

PR #39, opened 2025-04-11
Reverse-Stop + Error Pushback corner case fixed

surfdado:
Reverse stop detection now takes precendence over error condition detection, aka wheelslip or pushback.

Also, if we get error pushback (HV/LV/Temp) and then try to stop we need to take the already tilted nose into account to avoid briefly releveling the board.

Fix: don't ignore reverse-stop during error pushback situations

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -87,6 +87,7 @@ static void cmd_flywheel_toggle(Data *d, unsigned char *cfg, int len);
 
 const VESC_PIN beeper_pin = VESC_PIN_PPM;
 
+#define REVSTOP_ERPM_INCR 0.00008
 #define EXT_BEEPER_ON() VESC_IF->io_write(beeper_pin, 1)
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -519,6 +520,18 @@ static void calculate_setpoint_target(Data *d) {
                 }
             }
         }
+    } else if (d->float_conf.fault_reversestop_enabled && d->motor.erpm < -200 && !d->state.darkride) {
+        // Detecting reverse stop takes priority over any error condition SAT
+        if (d->state.sat >= SAT_PB_HIGH_VOLTAGE) {
+            // If this happens while in Error-Tiltback (LV/HV/TEMP) then we need to
+            // take the already existing setpoint into account
+            d->reverse_total_erpm =
+                -1 * (d->reverse_tolerance + d->setpoint_target_interpolated / REVSTOP_ERPM_INCR);
```
lukash, line 529: Nitpick: an unary `-` would be more elegant here.

PR #44, opened 2025-05-05
LEDs: provide colors for the status battery indicator

acheronfail:
I thought rather than just a white progress bar, that colours could be more visually intuitive at a glance.

Interested to hear your thoughts @lukash!

### DIFF
--- a/src/leds.c
+++ b/src/leds.c
@@ -76,11 +76,15 @@ static const uint32_t colors[] = {
     0x00FF70A0,  // LAVENDER
 };
 
-#define BATTERY_COLOR 0x00909090
+static const uint32_t battery_colors[] = {
+    0x00FF0000,
+    0x00FFB040,
+    0x0088FF00,
+};
+
 #define DUTY_COLOR 0x00FFB030
 #define FOOTPAD_SENSOR_COLOR 0x0000C0FF
 #define RED_BAR_COLOR 0x00FF3828
-#define BATTERY10_BAR_COLOR 0x00FF5038
 #define CONFIRM_COLOR 0x00A040FF
 
 #define CONFIRM_ANIMATION_DURATION 0.8f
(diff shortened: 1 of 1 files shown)


lukash:
I'm not a fan of bar color changing based on the value honestly. The yellow you used is very confusable with the duty cycle color and I'd much rather prefer to use other colors for representing different data rather than using up a lot of the colors just to express the battery level, which is already expressed by the bar, so the information is redundant. I prefer to "see white bar" -> "know it's battery percentage" (with the exception that it's red when below some point, I think that's fair enough for when it's really low).

An example of what I wanted to use other colors for is to have also the motor and battery current limits shown as part of the duty cycle bar, with different colors.

acheronfail:
All great points! I was just in the code and thought it might be nice, and was so easy to create a PR that I did that rather than an issue to discuss! :laughing: 

> An example of what I wanted to use other colors for is to have also the motor and battery current limits shown as part of the duty cycle bar, with different colors.

How would this work? How can you show more than a single bar at once?
Or would you make it configurable?

lukash:
> How would this work? How can you show more than a single bar at once?
> Or would you make it configurable?

I'd show the maximum of the three values, like FM now shows it in the app (AFAIK, haven't opened an FM app in more than a year), Floaty might have that too (I gotta check that one out now that I can without too much hassle). Alternatively, I could color all three of them one over the other (the longest one being at the bottom), that'll probably be too busy but could be cool, will probably at least try it.

I don't have time to mess with LEDs lately though :sweat_smile:

acheronfail:
@lukash that sounds like a cool idea.

When I have some time I'll see if I can get a PR up to do that - I'm more than happy to continue contributing LED effects.

lukash:
That'd be great! To give a few pointers, the current saturation is now returned by `motor_data_get_current_saturation()` and only the larger value is returned. This is called from haptic feedback. I'd instead store both current saturations into variables in `MotorData` and then pull those two along with the duty cycle from `MotorData` in LED code.

Totally unrelated, but if were interested: When the battery is close to 0%, the whole battery bar goes dark. Instead, when the last LED starts fading, it should instead fading start blinking and stay blinking even when at (or below) 0%.

PR #45, opened 2025-05-23
Ignore push start backwards if reverse stop is enabled

surfdado:
this prevents unintended re-engagements when trying to stop with rev-stop in some situations while push-start is also enabled

Fix: prevent unsafe situations with reverse stop + push start

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -172,6 +172,8 @@ static void configure(Data *d) {
 
     lcm_configure(&d->lcm, &d->float_conf.leds);
 
+    d->dt = 1.0f / d->float_conf.hertz;
+
     // Loop time in microseconds
     d->loop_time_us = 1e6 / d->float_conf.hertz;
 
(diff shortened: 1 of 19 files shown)


src/main.c
```diff
@@ -958,8 +958,12 @@ static void refloat_thd(void *arg) {
...
                     // 45 to prevent board engaging when upright or laying sideways
                     // 45 degree tolerance is more than plenty for tricks / extreme mounts
-                    engage(d);
-                    break;
+                    if (d->float_conf.fault_reversestop_enabled && d->motor.erpm < 0) {
+                        // ignore push start backwards if reverse stop is enabled
+                    } else {
+                        engage(d);
+                        break;
+                    }
```
lukash, line 966: There are already two nested `if`s, which are effectively a logical `and`. You're adding a third nested `if`, but this one with a twist of having an empty `if` branch and non-empty `else` branch. Could you please not unnecessarily increase complexity, and ideally reduce it? All the expressions are `and`ed, IMO they could be all grouped together into a single `if` with a line-break separated list of conditions, might take some fiddling with the formatter to also get the comments in...

NicoAleman:
Guessing this PR was replaced with this merged PR: https://github.com/lukash/refloat/pull/65

If so, I imagine this can be closed.

lukash:
Huh.....

PR #47, opened 2025-05-30
Rotate Gyro Y to estimate Pitch Delta

NicoAleman:
Fix: Rotate Gyro Y to estimate the change in pitch due to rider input, without influence from turning

### DIFF
--- a/src/imu.c
+++ b/src/imu.c
@@ -33,14 +33,22 @@ void imu_init(IMU *imu) {
 }
 
 void imu_update(IMU *imu, const BalanceFilterData *bf, const State *state) {
+    float roll_rad = VESC_IF->imu_get_roll();  // in Radians
+
     imu->pitch = rad2deg(VESC_IF->imu_get_pitch());
-    imu->roll = rad2deg(VESC_IF->imu_get_roll());
+    imu->roll = rad2deg(roll_rad);
     imu->yaw = rad2deg(VESC_IF->imu_get_yaw());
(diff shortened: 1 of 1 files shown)


src/imu.c
```diff
@@ -33,14 +33,21 @@ void imu_init(IMU *imu) {
...
     imu->balance_pitch = rad2deg(balance_filter_get_pitch(bf));
 
     float gyro[3];
     VESC_IF->imu_get_gyro(gyro);
-    imu->gyro_y = gyro[1];
+
+    float sin_roll = sinf(roll_rad);
+    float cos_roll = cosf(roll_rad);
+
+    // Rotated to estimate Pitch Rate due to Rider Input (Credit to Michalmo)
```
lukash: Code looks good, I just find the comment doesn't explain the idea behind the math and the "... due to Rider Input" part kind of looks like it might be expressing an important point but it's not clear what it is, I assume it's supposed to be in contrast to pitch rate introduced by the board turning...

I think the important thing is the math mixes in rotation in yaw the higher the roll is. Not sure the trigonometric functions can really be explained easily...

Also, not too important but since I'm commenting on this, the credits would be better in the commit message rather than the code comment IMO.

NicoAleman: Updated

michalmo:
Looking at this I think that I myself might not understand where this came from 2 years from now, so I'll add the derivation here for anyone trying to track this down in the future.

image

PR #50, opened 2025-06-17
Added tilt smoothing filters to Turn Tiltback

NicoAleman:
Feature: Added experimental setpoint filters to Turn Tiltback

### DIFF
--- a/src/atr.c
+++ b/src/atr.c
@@ -56,7 +56,7 @@ void atr_configure(ATR *atr, const RefloatConfig *config) {
     );
 }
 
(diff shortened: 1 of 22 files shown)


lukash:
I'll hold off with this for a bit if you don't mind, I've got a number of commits on top of the filtering, I need to consolidate them and try to move the whole thing forward. The way the filtering configuration is shared right now is clunky and we can now have more options (:sweat_smile:), so I'll try to come up with something better.

(It also conflicts now, but don't rebase just yet unless you want to, I'll be likely shuffling commits in `testing` soon, so it'll probably conflict again)

PR #51, opened 2025-06-18
Fixed logic for tilt/release speed conditions

NicoAleman:
Fix: Tilt Release Speed (TT/ATR) is no longer used when crossing 0°;
 only used when moving towards smaller angle of same sign or 0°

### DIFF
--- a/src/ema_filter.c
+++ b/src/ema_filter.c
@@ -51,10 +51,12 @@ void ema_filter_update(EMAFilter *filter, float target, float dt) {
     filter->speed += dt * filter->accel;
 
     uint16_t speed_limit;
-    if (sign(filter->speed) == sign(filter->value)) {
-        speed_limit = filter->on_speed;
-    } else {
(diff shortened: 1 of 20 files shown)


src/ema_filter.c
```diff
@@ -51,10 +51,12 @@ void ema_filter_update(EMAFilter *filter, float target, float dt) {
     filter->speed += dt * filter->accel;
 
     uint16_t speed_limit;
-    if (sign(filter->speed) == sign(filter->value)) {
-        speed_limit = filter->on_speed;
-    } else {
+    if ((filter->value * target >= 0) && (fabsf(filter->value) > fabsf(target))) {
```
lukash, line 54: Could you please remove the extra parentheses around the comparison expressions? They are redundant and this is a very common case where the operator precedence should be clear enough. I don't mind explicit parentheses in uncommon/unclear situations but here they IMO just add visual clutter.

lukash (commented):
`testing` has moved on since the PR was posted, could you please rebase?

Issue #46, opened 2025-05-25
Feature request: Speed/ERPM based pushback and buzz


buestad:
It would be really nice to have a speed based pushback and buzz implemented.

In many countries there is an upper speed limit for small electric vehicles. This also needs to be not controlled by app, so if possible this limit, for example 20km/h for my country, should not be possible to change in the VESC tool app, but only in the desktop version. Or even hard coded, so one could coose a 20km/h version of Refloat, and possibly other limits for other countries.

(I can participate in developing and testing. I work as a developer but have no coding experience in lisp)

acheronfail:
My two cents here:

I would love to have speed-based pushback/buzz, to limit the speed on the wheels when I let my kids/friends use the board.

That said, I disagree with the ability of not allowing the VESC Tool app to change it rather than the desktop app (I don't even think that's possible right now). Some owners don't have computers which can connect to the boards, and the mobile device is the only way.

VESC Tool is more of a configurator, you could argue that ride apps like Float Control or Floaty are the "apps" you use while riding - and they wouldn't be able to change this setting (if you really need to argue that).

lukash:
It's not (without ugly hacks on the VESC Tool side) possible to make something configurable only on the desktop version and not on mobile. Realistically it's not possible to make some configuration value "unchangeable" in the tool, the only realistic option for this I see is a different package with the value hard-coded. But, if these were to be in the package store, it would create somewhat of a mess in the package list for users and the build system with the repository is not made to handle this, not sure what that would take.

The code is C, no lisp knowledge needed.

FTR there's been requests for speed _notifications_ at certain speeds, which is different from a pushback (+ haptic alert?).

buestad:
I have made an attempt on developing this. I have a working copy where I can set a speed limit in the pushback section and have haptic feedback. I've tested this on my Pint which runs VESC 6.05 at low speeds and it works. If anyone are willing to test I can send the bin file, or you can clone my fork of Refloat and build yourself. Use the speed_pushback branch.

![Image](https://github.com/user-attachments/assets/860f81eb-b3b6-4bfe-aca7-6e02db00d551)

acheronfail:
I’ll definitely try this out sometime this week - thanks for that. I’m happy to clone and build https://github.com/buestad/refloat/tree/speed_pushback myself. 👍

buestad:
[...]

PR #52, opened 2025-06-18
Board Speed: Implemented pushback and haptic feedback

buestad:
Feature: Added functionality for speed-based pushback and haptic buzz. Configurable in Refloat Cfg

https://github.com/lukash/refloat/issues/46


lukash:
Thanks for the PR, skimmed through, looks pretty clean. I'll have to think about the new config options (why not use the same settings as for duty pushback?) and the new pushback SAT (here a specific SAT might be the way to go, but in general I wanna make one SAT be "alert pushback" and then have another enum as the alert reason).

The CI has failed on `clang-format`, there's a documentation on how to run it in the cotribution guidelines. I'll be busy the next week and a bit, might be slow to response...

buestad:
I Tried running the clang-format (also throug lefthook) but it fails:

```
$ clang-format -i src/main.c 
.../refloat/.clang-format:7:24: error: invalid boolean
AlignTrailingComments: Never
                       ^~~~~
```

buestad:
About the config options, It might make sense to use the same angle and angle-speed as duty, but its nice to have a different haptic tone. I don't wanna ignore duty, but speed, not so important :)

lukash:
> I Tried running the clang-format (also throug lefthook) but it fails:
> 
> ```
> $ clang-format -i src/main.c 
> .../refloat/.clang-format:7:24: error: invalid boolean
> AlignTrailingComments: Never
>                        ^~~~~
> ```

Looks like you have an old version of clang-format, you need version 18.

buestad:
> Looks like you have an old version of clang-format, you need version 18.

Fixed

lukash:
@buestad can you please squash the clang-format fix with the main commit and also drop the merge commits? Instead (if needed) please _rebase_ the branch onto the current `main`.

lukash:
Also, can you just globally replace `board_speed` for just `speed` and analogously for other ways of writing it (SCREAMING_CASE, CamelCase and whatnot) :grin:? The word "board" is redundant...

And, in general the goal is to keep the config options minimal necessary and I think we can use the duty settings for mostly everything, the only new option should be the actual speed setting. The options for duty cycle which will now also be used for speed alerting just need to be updated (names and descriptions) to say they now apply to Speed too.

src/main.c
```diff
@@ -644,6 +648,10 @@ static void calculate_setpoint_target(Data *d) {
             d->state.sat = SAT_NONE;
             d->setpoint_target = 0;
         }
+    } else if (d->float_conf.tiltback_board_speed > 0.0 && d->motor.speed > d->float_conf.tiltback_board_speed) {
```
lukash: The `d->motor.speed > d->float_conf.tiltback_board_speed` part of the condition should take absolute value to apply also when going battery forward.

src/main.c
```diff
@@ -644,6 +648,10 @@ static void calculate_setpoint_target(Data *d) {
...
+        d->beep_reason = BEEP_BOARD_SPEED;
+        d->setpoint_target = d->float_conf.tiltback_board_speed_angle;
```
lukash: ... and here you need to do the same thing as in the other branches, depending on the direction make the setpoint target negative, to work also when going battery forward.

buestad:
I can do the fixes tomorrow 👍 About board_speed. The reason for choosing that was because some other variables in the same area that had speed in the name. Believe it was related to the speed of pushback or something.

src/conf/datatypes.h
```diff
@@ -218,6 +218,7 @@ typedef struct {
     float tiltback_lv_angle;
     float tiltback_lv_speed;
     float tiltback_lv;
+    float tiltback_speed;
```
lukash: Nitpick, can you please move this below `tiltback_duty` a few lines above, just so that the related options are together?

src/conf/settings.xml
```diff
@@ -733,6 +733,32 @@ p, li { white-space: pre-wrap; }
             <suffix> V</suffix>
             <vTx>7</vTx>
         </tiltback_lv>
+        <tiltback_speed>
+            <longName>Board Speed Limit</longName>
```
lukash: "Speed Threshold" to keep it consistent with "Duty Cycle Threshold".

src/conf/settings.xml
```diff
@@ -733,6 +733,32 @@ p, li { white-space: pre-wrap; }
...
+            <editAsPercentage>0</editAsPercentage>
+            <maxDouble>100</maxDouble>
+            <minDouble>0</minDouble>
+            <showDisplay>0</showDisplay>
+            <stepDouble>1.0</stepDouble>
+            <valDouble>0.0</valDouble>
+            <vTxDoubleScale>1000</vTxDoubleScale>
+            <suffix> km/h</suffix>
+            <vTx>7</vTx>
+        </tiltback_speed>
```
lukash, line 946: This option can be of type Int, TX Type UINT_8, that saves on the total config size. And the step would be better at 5.

At a later point I'll edit the option description a bit myself, it'll be easier than commenting on the bloated HTML text...

src/conf/settings.xml
```diff
@@ -871,7 +897,7 @@ p, li { white-space: pre-wrap; }
 &lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
 p, li { white-space: pre-wrap; }
 &lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'Roboto'; ; font-weight:400; font-style:normal;&quot;&gt;
-&lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;Audible frequency of the Duty Cycle haptic feedback. Also used for Current Limit haptic Feedback, if enabled.&lt;/p&gt;
+&lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;Audible frequency of the Duty Cycle haptic feedback. Also used for Current Limit and Board Speed haptic Feedback, if enabled.&lt;/p&gt;
```
lukash: Instead of "Board Speed" I'd use "Speed alert" here, ditto on the strength right below.

src/haptic_feedback.c
```diff
@@ -83,6 +85,8 @@ static uint8_t get_beats(HapticFeedbackType type) {
         return 2;
     case HAPTIC_FEEDBACK_DUTY_CONTINUOUS:
         return 0;
+    case HAPTIC_FEEDBACK_SPEED:
+        return 3;
```
lukash: I'm not sure off the bat what tone pattern this does, but here I'd keep it simple and use the same pattern as duty. That means you don't need the extra `HAPTIC_FEEDBACK_SPEED`, you can reuse `HAPTIC_FEEDBACK_DUTY` and ideally rename it to something like `HAPTIC_FEEDBACK_SPEED_DUTY`.

src/conf/settings.xml
```diff
@@ -3968,6 +3995,8 @@ p, li { white-space: pre-wrap; }
                     <param>tiltback_duty_angle</param>
                     <param>tiltback_duty_speed</param>
                     <param>is_dutybeep_enabled</param>
+                    <param>::sep::Board Speed Alert Pushback</param>
```
lukash: "Speed Alert Pushback" here I guess. As I said the options need some restructuring, there's not much point to have a section with only one item, ...

src/haptic_feedback.c
buestad: Decided to keep `HAPTIC_FEEDBACK_SPEED`. but tidy up the cases in stead. Felt kinda messy to use a common one because of the `HAPTIC_FEEDBACK_DUTY_CONTINUOUS`

src/conf/settings.xml
buestad: fixed

src/conf/settings.xml
buestad: fixed

src/conf/settings.xml
buestad: fixed

src/conf/settings.xml
buestad: fixed

src/conf/datatypes.h
buestad: fixed

src/conf/settings.xml
lukash: The option is still of type Double and step is still 1...

src/haptic_feedback.c
lukash: But it's now redundant, it represents the same thing as `HAPTIC_FEEDBACK_DUTY`, please unify it into one item...

src/conf/settings.xml
buestad: Oh. I misundestood. I thaugt that vTx = 5 ment double. Should vTx be 5 or 7?  What is vTx? You think that step size in km/h should be 5? Isn't that a bit course?

buestad: fixed. Used VESC param editor to review and change the XML

src/haptic_feedback.c
buestad: Changed `HAPTIC_FEEDBACK_DUTY` to `HAPTIC_FEEDBACK_DEFAULT` and adopted

src/haptic_feedback.c
```diff
@@ -55,8 +55,10 @@ static HapticFeedbackType haptic_feedback_get_type(
         if (md->duty_cycle > hf->duty_solid_threshold) {
             return HAPTIC_FEEDBACK_DUTY_CONTINUOUS;
         } else {
-            return HAPTIC_FEEDBACK_DUTY;
+            return HAPTIC_FEEDBACK_DEFAULT;
```
lukash: While I don't think `HAPTIC_FEEDBACK_SPEED_DUTY` is a great name, I think it's still better than `HAPTIC_FEEDBACK_DEFAULT`, which carries no information about what it represents... Could you use that? (or provide arguments why yours is better :grin:)

buestad: I think that default is better because it is general. Now this goes for two or three things (duty, speed and maybe current?) But I think maybe we can do a refactor later if needed. Changed for now to `HAPTIC_FEEDBACK_SPEED_DUTY ` and we can discuss in Discord :)

lukash: The current is using `DUTY_CONTINUOUS` (not reflected in the name) :sweat_smile:

lukash:
I changed the target branch for the PR from `main` to `testing`, we normally do this for features that should get some testing (and I move them to `main` manually later on).

acheronfail:
Just wanted to stop by and mention I’ve been running this with a speed limit of 15kmh on a PintV that my son’s been using for 2 weeks and have had no issues whatsoever so far 👍

PR #58, opened 2025-09-22
Correct UI typo in settings.xml

OakCityDirtyRandy:
Fix three instances of a recurring minor typo that appears in the UI.  "Tilitback" should be "Tiltback".


lukash:
Thanks!

PR #60, opened 2025-10-09
Added Tilt Smoothing to TurnTilt + BrakeTilt

NicoAleman:
Feature: Added Tilt Smoothing to Turn Tiltback and Brake Tiltback


lukash:
Same note about imperative instead of past tense in the changelog message applies (i.e. use _"Feature: Add Tilt Smoothing ..."_ instead of _"Feature: Added Tilt Smoothing ..."_).

PR #61, opened 2025-10-09
Added hundredths precision to Brake Scaling params

NicoAleman:
Feature: Added hundredths precision to Brake Scaling params

Myself and others I've seen have encountered a few scenarios where the current tenths place precision wasn't as high resolution as desired. 

Honestly ideally I feel like it'd make more sense at some point to transition to just separate Angle/Rate P values for braking (maybe setting to 0 would just use same values as normal Angle/Rate P) instead of scaling where it's not easy to change just the nose behavior without the brakes also being affected. But that's another convo, and higher precision would at least make it easier to mimic specific desired Angle/Rate P values for just the brakes


lukash:
Again, just the changelog record imperative instead of past tense...

PR #63, opened 2025-10-10
Reworked ATR Tilt Speed conditions

NicoAleman:
Fix: Reworked ATR Tilt Speed conditions:
 - Re-organized implementation to avoid repeated logic
 - Accounts for ATR Thresholds in the Boost margin
 - Uses "On" Tilt Speed for Transition Boost if faster

The margins originally in place to determine whether or not to use Transition Boost were before ATR Thresholds, when ATR Setpoint would constantly jump around between positive and negative when cruising. With these thresholds in place, the margins are less necessary and achieve the same effect. This fix negates the thresholds from the margins in effect, making the response the same as originally intended (unless the thresholds are larger than the original margins, then those are respected).

Additionally, depending on configuration and your current speed, there can be situations where Tilt Speed * the current Response Boost is faster than the Transition Boost speed. So for transition boost, we look at both tilt speeds and use the fastest one available, similarly to how transitions are now handled in Torque Tiltback in my other Pull Request.

The logic was also just reorganized as a whole, the original structure had unnecessary repeating logic and was slightly hard to follow. It's still a lot here thanks to the new logic, but I consolidated what I can.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: cachix/install-nix-action@v25
       - uses: rrbutani/use-nix-shell-action@v1
         with:
-          flakes: nixpkgs#clang-tools_18, nixpkgs#lefthook
+          flakes: nixpkgs#llvmPackages_18.clang-tools, nixpkgs#lefthook
 
       - name: Run clang-format
         id: clang-format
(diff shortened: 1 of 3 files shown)


lukash:
@NicoAleman it seems your commit should be split into three commits, according to the bullet points in your commit message. Can you please do that? That way it's way easier to understand the changes. I also came to look at the simplification you did, as I'm making other changes and I thought I'd pull that in right away to minimize conflicts, but like this I can't do it.

(Also the code cleanup changes don't need to go to the changelog)

NicoAleman:
@lukash Let me know if there's any issue. Had to update the CI file to pass checks, not sure if appropriate to include in this PR but wasn't sure how best to go about it otherwise.

lukash:
This has also been reworked for 1.3 and the changes are no longer relevant, closing...

PR #66, opened 2025-10-20
Turntilt fix: don't suppress it in high torque response situations

surfdado:
Originally I had added this turntilt suppression beacuse of my paranoia about getting taildrags on downhills when turntilt kicks in...

It turns out that it was unnecessary. In the contrary: uphill/downhill berms, no matter how steep, still benefit from having turntilt.

Fix: remove turntilt suppression in high torque response situations


lukash:
@surfdado nitpick, could you lead the changelog records with a capital letter? I'm fixing these manually in the changelog...

Also, conflicts now.

surfdado:
Should be all good now

PR #71, opened 2026-01-29
Fix unintended extra short beep on package start

0x41kravchenko:
While using beeper on servo (PPM) pin I noticed that my board makes 2 short beeps and 1 long beep on startup.
And knowing this [On boot: short beep when VESC boots, long beep when balance app is ready (and each time you write app config!) or triple-beep when float package is disabled](https://pev.dev/t/buzzer-for-float-package/99) from Float package article by surfdado (IFAIK beeper code was unchanged in refloat) it made me think that 1 extra short beep is due to some errors/faults.
According to current implementation correct behaviour indeed should be 1 short beep (in `configure()`) and 1 long beep (when board is ready).
1 short beep (in `configure()`):
https://github.com/lukash/refloat/blob/a86a3eb3c64a3672bca154c6f03922f5b218750b/src/main.c#L227
1 long beep (when board is ready)
https://github.com/lukash/refloat/blob/a86a3eb3c64a3672bca154c6f03922f5b218750b/src/main.c#L847-L848

That one extra short beep happens unintentionally due to beeper_pin having high level right after pin init here:
https://github.com/lukash/refloat/blob/a86a3eb3c64a3672bca154c6f03922f5b218750b/src/main.c#L105
And since `beeper_update` waits for `beep_countdown` to be 0 and does nothing to the pin after `beep_alert()` the pin remains high for that first beep_countdown period which makes that beep.

Also I wanted to try to build the package on MacOS which was successful but since vesc_tool path in MacOS contains spaces I had to wrap `$(VESC_TOOL)` in double quotes in Makefile. I double checked this change on Linux and it is not causing any issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,10 @@ If you don't have `vesc_tool` in your `$PATH` (but you have, for example, a down
 ```sh
 make VESC_TOOL=/path/to/vesc_tool
 ```
+For macOS, the path to VESC Tool when installed using the official installer is as follows:
+```sh
+make VESC_TOOL="/Applications/VESC Tool.app/Contents/MacOS/VESC Tool"
+```
 
 ## Documentation
 [Development Documentation](doc/index.md)
--- a/Makefile
+++ b/Makefile
@@ -9,9 +9,9 @@ all: refloat.vescpkg
 
 refloat.vescpkg: src lisp/package.lisp package_README-gen.md ui.qml
 ifeq ($(OLDVT), 1)
-	$(VESC_TOOL) --buildPkg "refloat.vescpkg:lisp/package.lisp:ui.qml:0:package_README-gen.md:Refloat"
+	"$(VESC_TOOL)" --buildPkg "refloat.vescpkg:lisp/package.lisp:ui.qml:0:package_README-gen.md:Refloat"
 else
-	$(VESC_TOOL) --buildPkgFromDesc pkgdesc.qml
+	"$(VESC_TOOL)" --buildPkgFromDesc pkgdesc.qml
 endif
 
 src:
--- a/src/main.c
+++ b/src/main.c
@@ -103,6 +103,7 @@ const VESC_PIN beeper_pin = VESC_PIN_PPM;
 
 void beeper_init() {
     VESC_IF->io_set_mode(beeper_pin, VESC_PIN_MODE_OUTPUT);
+    VESC_IF->io_write(beeper_pin, 0);
 }
 
 void beeper_update(Data *d) {
@@ -1232,6 +1233,10 @@ static void data_init(Data *d) {
     );
 
     d->odometer = VESC_IF->mc_get_odometer();
+
+    d->beep_num_left = 0;
+    d->beep_duration = 0;
+    d->beep_countdown = 0;
 }
 
 // See also:


README.md
```diff
@@ -30,6 +30,10 @@ If you don't have `vesc_tool` in your `$PATH` (but you have, for example, a down
 ```sh
 make VESC_TOOL=/path/to/vesc_tool
 ```
+macOS:
+```sh
+make VESC_TOOL=/Applications/VESC\ Tool.app/Contents/MacOS/VESC\ Tool
```
lukash: Is this always the path where VESC Tool is installed on MacOS? The command right above it is generic and applies to Mac as well, you're adding a special case of it. I'm not too fond of adding the special case, if someone is compiling the package he's probably savvy enough to know the path and the generic instruction should be enough.

If the path on MacOS is standardized and always the same, I'm not entirely against it but it should be labeled more specifically as a special case of the above command for MacOS, e.g.: _"For macOS, the path to VESC Tool is as follows:"_

I'd also slightly favor quoting the string here instead of ascaping the spaces:
```
make VESC_TOOL="/Applications/VESC Tool.app/Contents/MacOS/VESC Tool"
```

lukash (commented):
Thanks! Looks good, just one small nitpick.

lukash:
Oh, for the beeper fix, please add a changelog record via a git trailer as described in https://github.com/lukash/refloat/blob/main/CONTRIBUTING.md

You can also find examples earlier commits.

README.md
0x41kravchenko: Thank you for the feedback!
Yes, it is default path if VESC tool is installed with MacOS package from official VESC website. Of course if one would compile VESC Tool then the path can be different.
I like the idea with a special case label, also it will be an indicator for other contributors that building packages with  MacOS VESC Tool is doable. I'll update README according to your example.
At first I didn't want to update the Makefile and tried quoting the path in cli but command substitution in Makefile still interprets path with spaces as multiple arguments. Here's a test with reverted change:
```
% git diff
diff --git a/Makefile b/Makefile
index 411013d..b069eff 100644
--- a/Makefile
+++ b/Makefile
@@ -9,9 +9,9 @@ all: refloat.vescpkg

 refloat.vescpkg: src lisp/package.lisp package_README-gen.md ui.qml
 ifeq ($(OLDVT), 1)
-       "$(VESC_TOOL)" --buildPkg "refloat.vescpkg:lisp/package.lisp:ui.qml:0:package_README-gen.md:Refloat"
+       $(VESC_TOOL) --buildPkg "refloat.vescpkg:lisp/package.lisp:ui.qml:0:package_README-gen.md:Refloat"
 else
-       "$(VESC_TOOL)" --buildPkgFromDesc pkgdesc.qml
+       $(VESC_TOOL) --buildPkgFromDesc pkgdesc.qml
 endif

 src:
 
% make VESC_TOOL="/Applications/VESC Tool.app/Contents/MacOS/VESC Tool"
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C src
arm-none-eabi-gcc alert_tracker.so atr.so balance_filter.so biquad.so bms.so booster.so brake_tilt.so charging.so data_recorder.so footpad_sensor.so haptic_feedback.so imu.so konami.so lcm.so led_driver.so led_strip.so leds.so main.so motor_control.so motor_data.so pid.so remote.so state.so time.so torque_tilt.so turn_tilt.so utils.so lib/circular_buffer.so conf/confparser.so conf/confxml.so conf/buffer.so ../vesc_pkg_lib//utils//rb.so ../vesc_pkg_lib//utils//utils.so -nostartfiles -static -mfloat-abi=hard -mfpu=fpv4-sp-d16 -mcpu=cortex-m4 -lm -Wl,--gc-sections,--undefined=init -T ../vesc_pkg_lib//link.ld -flto -o package_lib.elf
lto-wrapper: warning: using serial compilation of 2 LTRANS jobs
lto-wrapper: note: see the '-flto' option documentation for more information
/Applications/ArmGNUToolchain/14.2.rel1/arm-none-eabi/bin/../lib/gcc/arm-none-eabi/14.2.1/../../../../arm-none-eabi/bin/ld: warning: package_lib.elf has a LOAD segment with RWX permissions
arm-none-eabi-objdump -D package_lib.elf > package_lib.list
arm-none-eabi-objcopy -O binary package_lib.elf package_lib.bin --gap-fill 0x00
python3 ../vesc_pkg_lib//conv.py -f package_lib.bin -n package_lib > package_lib.lisp
/Applications/VESC Tool.app/Contents/MacOS/VESC Tool --buildPkgFromDesc pkgdesc.qml
make: /Applications/VESC: No such file or directory
make: *** [refloat.vescpkg] Error 1
```

0x41kravchenko:
@lukash Updated the PR according to the nitpicks 👌

README.md
lukash: Nono, I didn't mean to remove the quotes in the Makefile (those are needed), I meant to put this into the README:
```sh
make VESC_TOOL="/Applications/VESC Tool.app/Contents/MacOS/VESC Tool"
```

That is, quotes instead of the backslashes. This is more conventional and it's easier to change the path to another path with spaces, so it's better to have it that way for documentation.

0x41kravchenko: Ah, I got your point now! Updated ✅

lukash:
Thank you!

lukash:
@0x41kravchenko hey, not sure this was doing anything for you but doesn't seem it really fixed the extra beep. Having a second look, the initialization of variables wasn't really making a change because there's a `memset(d, 0, sizeof(Data))` at the beginning of `data_init()` (though I do prefer explicit initialization at least in theory). The initial setting of the pin to off also shouldn't be doing anything if the pin is already off (if it wasn't it'd be a problem anyway).

The real reason of the extra beep was the first short beep was the initial `configure()` call. Then it was a race, if the IMU was initialized before the beep ended, it wouldn't do the long startup beep, if the first beep was finished, it'd do it.

0x41kravchenko:
Hey @lukash!
Just to confirm, do you still have 2 short beeps and 1 long beep on startup after the fix was applied?

I added variables initialisation for the beeper during investigation but it wasn't the root cause, so I decided to leave those for consistency as all other variables were initialised explicitly.

The real root cause of the extra short beep (the one before `configure()`) was beeper pin having high level after IO init (confirmed by commenting out `beeper_update()` in the loop and beeper is constantly ON after board startup) and the way `beeper_update()` works.
After beeper pin IO init it had high level and `beeper_update()` allowed that high level to persist for first `beep_countdown` period which doesn't do anything to beeper pin due to `if (d->beep_countdown <= 0)` is `false` for that first `beep_countdown` period.

Here's step by step first beep flow how I understood it (before the fix):
1) Beeper init in `INIT_FUN()` **(here it starts unintended beep due to pin's high level)** and then refloat thread is spawn
2) In the thread `configure()` is executed where it loads beeper vars with first intended beep values in `beep_alert(d, 1, false);` (beep_num_left = 1*2+1 = 3, beep_duration = 80, beep_countdown = 80)
3) Main loop starts and triggers `beeper_update()` on each iteration.
4) For 80 loop iterations it just decrements d->beep_countdown until `if (d->beep_countdown <= 0)` is true
5) On 81st iteration it resets beep_countdown back to 80 and decrement beep_num_left from 3 to 2, `if (d->beep_num_left & 0x1)` is false and beeper pin is set to low (**EXT_BEEPER_OFF()**) - **here the unintended beep ends**
6) beeper_update() again waits 80 iterations and then resets beep_countdown back to 80 and decrements beep_num_left from 2 to 1 and the beeper is ON (intended first beep starts)
7) 80 iterations wait and decrement beep_num_left from 1 to 0 and turn beeper OFF (intended first beep ends)
8) The above happens before it reaches `STATE_STARTUP`. After it reaches `STATE_STARTUP` the long beep is triggered.

Also wanted to note that I have CheapFOCer 2 V1 VESC and I have checked it's [hwconf](https://github.com/vedderb/bldc/blob/f20e89217810bce8625c69304e73dc30d3beb1eb/hwconf/shaman/Cheap_FOCer_2/hw_Cheap_FOCer_2_core.c) to be sure that servo pin is not set to high there and I think it's not the case.

I have captured servo pin state after startup with an oscilloscope
Before the fix:
photo_2026-03-09_12-24-14

After the fix:
photo_2026-03-09_12-24-19

lukash:
@0x41kravchenko thanks for that extremely detailed writeup! Ok so two separate issues here apparently. I was testing this on a Thor where the beeper pin is already initialized and used in hwconf (to signal the board being turned on at early boot stages). I was also doubting the immediate setting of the pin to low, because I thought if it'd be high before the package init, the beeper would be beeping. But apparently at least for your controller the pin was not configured as output, so it only started beeping once we configured it so (and didn't set the state). So yeah, you're fixing this with your change, apologies for doubting that :grin: .

I was fixing another issue, you can check out the commits in `main`. The first short configure beep was non-deterministically preventing the long beep of the package ready signal, depending on whether the short beep ended before the IMU was initialized or not (because the long beep would be dropped if the short one wasn't finished). I was made aware of this second beep and assumed your fix was supposed to address this, when it didn't, I dug into it and for a while thought it was just your fix missing the mark :joy: Once again, apologies and thanks for the contribution!

0x41kravchenko:
No worries, now I see where our confusion come from 🙂
Interesting to know that some hwconf's have beeper/servo init and that those controllers didn't have this issue in the first place.

I checked the commits and I understand that another issue you fixed. I have never noticed such behaviour, maybe IMU init on my controller takes longer allowing first beep to finish before second (long) one.

Thank you for the explanation!

PR #73, opened 2026-02-24
Fix: haptic strength scaling is now bidirectional

surfdado:


lukash:
I'm sorry but the changelog message is really unclear, could you phrase it like this?
> Fix: Fix haptic strength speed scaling being applied in reverse when going backwards

PR #74, opened 2026-02-26
Remove footpads disabling LEDs on startup

Eli-Parker:
Remove the disabling of LEDs if both sensors are pressed during startup. Settings.xml LED Mode description is updated to reflect the change

### DIFF
--- a/src/conf/settings.xml
+++ b/src/conf/settings.xml
@@ -3591,8 +3591,7 @@ p, li { white-space: pre-wrap; }
 &lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;The rest of the options is ignored for the &lt;span style=&quot; font-style:italic;&quot;&gt;External&lt;/span&gt; LED type.&lt;/p&gt;
 &lt;p style=&quot;-qt-paragraph-type:empty; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;br /&gt;&lt;/p&gt;
 &lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;Board restart required for changes to take effect.&lt;/p&gt;
-&lt;p style=&quot;-qt-paragraph-type:empty; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;br /&gt;&lt;/p&gt;
-&lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;Note: Pressing both footpad sensors when turning on the board will disable the LED initialization. Use if the LEDs break your VESC in a way that prevents you to connect and change configuration (e.g. due to an overcurrent).&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</description>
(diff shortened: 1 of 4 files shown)


src/conf/settings.xml
```diff
@@ -3591,8 +3591,7 @@ p, li { white-space: pre-wrap; }
...
 &lt;p style=&quot;-qt-paragraph-type:empty; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;br /&gt;&lt;/p&gt;
 &lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;Board restart required for changes to take effect.&lt;/p&gt;
-&lt;p style=&quot;-qt-paragraph-type:empty; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;br /&gt;&lt;/p&gt;
-&lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;Note: Pressing both footpad sensors when turning on the board will disable the LED initialization. Use if the LEDs break your VESC in a way that prevents you to connect and change configuration (e.g. due to an overcurrent).&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</description>
+&lt;p style=&quot;-qt-paragraph-type:empty; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;br /&gt;&lt;/p&gt;</description>
```
lukash: You used a text editor for `settings.xml`, right? Unfortunately editing the descriptions sucks.

There's a GUI editor in VESC Tool, in the top menu go Developer -> Parameter Editor Mcconf (yeah, Mcconf, you can probably use any of them). Then open the XML, you can edit it and save it...

The point is the html descriptions are uneditable by hand. If you use the editor it'll always reset the html formatting and so if you make changes by hand that are different from the generated thing, next time someone uses the editor, they'll have an extra diff. So we're effectively forced to use it.

I normally add items in the XML itself and edit partially by hand (the trivial/obvious things). For changing descriptions or the enums etc. I use the GUI thing.

I tried loading and saving the XML and I did get a diff, also (nitpick) there was an extra linebreak at the end of the description.

lukash (changes requested):
Thank you, code LGTM.

I'm using a git commit trailer mechanism for tracking changelog records, could you add such changelog record to the commit message? It'll also automatically attribute the change to the commit author name in the changelog. Just add a "Feature: ..." trailer, it's described in `CONTRIBUTING.md`, you can also check past commits for examples and you can run `./changelog.py` to see what it outputs. (note I use a very crude mechanism for multiline formatted content of the git trailer, not proud of it but it does the job...)

My idea of what it should look like (feel free to improve it):
```
Feature: Don't disable LEDs if both sensors are pressed at startup
 >
 This mechanism was in place for the potential case of the LED
 initialization locking up the board in a way that would prevent the
 user from connecting and fixing the problem. This has proven unneeded,
 so the feature is being removed.
```

Eli-Parker:
I've amended with the feature tag added and the updated XML from VESC Tool. All looks good in changelog.py.

src/conf/settings.xml
Eli-Parker: Should be taken care of now! I'll keep this in mind for all future edits.

lukash:
@Eli-Parker I think you somehow managed to drop the code changes from the commit :sweat_smile: 

settings.xml looks good now, in the commit message ideally the _"Remove the disabling of LEDs if both sensors are pressed during startup. Settings.xml LED Mode description is updated to reflect the change"_ line should be wrapped as it's too long (and it's missing a period :grin: ).

Eli-Parker:
My bad, it must've gotten dropped in the amend

Eli-Parker:
I've verified that all the changes are back onto the commit. I seem to be having a few problems getting changelog.py to run, something about an invalid escape sequence '\.'. Let me know if you notice anything else, sorry for the trouble with git.

lukash:
@Eli-Parker you've got a merge commit there now though, we don't want that... hit me up on discord if you need help getting rid of it.

Eli-Parker:
Got it, just needed to rebase.

lukash:
Looking good now, thank you!

PR #77, opened 2026-03-09
Fix disappearing last beep reason

Eli-Parker:
Fix: Stop the "last beep reason" text from disappearing once the beep condition stops

Got this fix finalized. I was also thinking the "sensor stop condition" and "last beep reason" texts should always be visible to let the users know that they're able to see that data if it's applicable.


lukash:
Nice catch with the FA calling the command. Could you just add " ... if Float Accessories is installed" to the Fix: trailer to make it a bit more specific?

Eli-Parker:
Should be all good!

PR #80, opened 2026-04-18
Add Nix support for building refloat, and use it in CI/CD

MilesBreslin:
Add Nix support for building refloat, and use it in CI/CD.

Nix builds of Refloat should be binary reproducible allowing auditing of CI/CD flows.

Notes for review:
- vesc-tool is pinned at `release_6_06`
- ~~nixpkgs is pinned at the lock-file value from vesc-tool@`release_6_06`~~
  - nixpkgs is now tracking `nixos/25.11`. The exact revision is whatever is latest on that branch at the time of my commit.
- `nix/refloat.nix` can be built standalone, but will likely need a `src` override and/or `GIT_HASH` override. `flake.nix` does this automatically.
- `gcc-arm-embedded` is pinned at `13` by `overlay.nix`. This resolves to `13.3.rel1`.
- Only `refloat.elf` and `refloat.vescpkg` are supported as outputs.
- Only `x86_64-linux` is supported because VESC Tool only supports `x86_64`.
- Updating any dependency version now requires updating `flake.lock` via nix.

I've included a full migration to Nix in this MR, but am happy to drop commits if this feels like an overreach. I personally want nix support to be able to develop this locally.

I have tested creating a release with this GitHub action in my own repo. It did not actually create a release, but I think that's a configuration issue for my Fork repo. There were no errors displayed when creating the release.
https://github.com/MilesBreslin/refloat/actions/runs/24614823553

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,45 @@
+{
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-25.11";
+    flake-utils.url = "github:numtide/flake-utils";
+    vesc-tool.url = "github:vedderb/vesc_tool/release_6_06";
+    vesc-tool.inputs.nixpkgs.follows = "nixpkgs";
+  };
+
+  outputs = {
+    self,
+    flake-utils,
+    nixpkgs,
+    vesc-tool,
+  } @ inputs:
+    flake-utils.lib.eachSystem ["x86_64-linux"] (
+      system: let
+        pkgs = import nixpkgs {
+          inherit system;
+          overlays = [
+            (final: prev: {
+              vesc-tool = vesc-tool.packages.${system}.vesc-tool;
+            })
+            (import ./nix/overlay.nix)
+            (final: prev: {
+              refloat = prev.refloat.overrideDerivation (old: {
+                version = prev.lib.strings.removeSuffix "\n" (builtins.readFile ./version);
+                src = self;
+
+                makeFlags = [
+                  "GIT_HASH=${builtins.substring 0 8 (self.rev or self.dirtyRev)}"
+                  "BUILD_DATE=${self.lastModifiedDate}"
+                ];
+              });
+            })
+          ];
+        };
+      in {
+        packages =
+          pkgs
+          // {
+            default = pkgs.refloat;
+          };
+      }
+    );
+}
--- a/nix/refloat.nix
+++ b/nix/refloat.nix
@@ -1,0 +1,34 @@
+{
+  lib,
+  stdenv,
+  gcc-arm-embedded,
+  vesc-tool,
+  python3,
+}:
+stdenv.mkDerivation {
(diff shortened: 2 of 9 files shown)


lukash:
Hello, thanks, some good things here for sure but I'll need to find some time to dive into it and figure out if it all lines up... Not sure when that's gonna be.

flake.nix
```diff
@@ -0,0 +1,45 @@
+{
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/7848cd8c982f7740edf76ddb3b43d234cb80fc4d"; # nixos-25.05 from vesc-tool lockfile
```
lukash: Just skimming through though this is not good. Not sure why exactly it's needed but I don't want a tie like this unless there are very good benefits that outweigh the drawbacks. Also no idea how well nix is maintained in the VESC Tool repo, I would assume Vedder doesn't actually care about it that much.

nix/refloat.nix
```diff
@@ -0,0 +1,34 @@
...
+}:
+stdenv.mkDerivation {
+  pname = "refloat";
+  version = "0.0.0-dev";
+
+  src = ../.;
+
+  postPatch = ''
+    # Python shebang is failing for some reason. Replace it with full shebang
+    sed -i '1s,.*,#!${lib.getExe python3},' rjsmin.py
```
lukash, line 16: This is ugly.

nix/refloat.nix
```diff
@@ -0,0 +1,34 @@
...
+    echo "$version" > ./version
```
lukash, line 17: There should ideally be one source of truth for version and you're duplicating it here. And version is set to 0.0.0-dev above. Not sure what this is supposed to be doing but given version is set to 0.0.0-dev above, not sure it even does what it's supposed to.

MilesBreslin: This delegates the single source of truth into nix, which defaults to reading the version file. If the nix package states a version, that will be embedded into the binary.

I designed `nix/refloat.nix` to be generic, and not dependent on the nix flake. `0.0.0-dev` is used as a placeholder in the generic build implementation. It is then overridden in the flake.nix by reading the `version` file. This keeps `nix/refloat.nix` generic, while maintaining expected behavior for development using the nix flake.

I think `echo "$version" > ./version` should stay, but if we want to change the placeholder version or shift where that logic is, no problem.

nix/refloat.nix
MilesBreslin: Yup. I tried getting it to work with `/usr/bin/env`, but it wasn't working when I tried it. I can take another look and see if there's a better way around this.

flake.nix
MilesBreslin: Amended. This points to `nixos-25.11` now, the latest available NixOS release. Note that 26.05 should be released towards the beginning of next month.

PR #82, opened 2026-04-28
Added motor and current saturation to the led status bar

acheronfail:
Following up on this suggestion here: https://github.com/lukash/refloat/pull/44#issuecomment-2866685440

A couple things to align on before this is ready to merge:
- [x] right now DUTY = orange, MOTOR I = magenta, BATTERY I = green - let me know what you'd prefer
- [x] currently when the battery is low the last LED pulses every second - do we want this faster/slower?
- [x] ~~should I split out the battery blinking into a separate commit + `feature:` / leave it as is / update the current commit message + trailer?~~ done in https://github.com/lukash/refloat/pull/84
- [x] ~~I renamed `status_duty_blend -> status_blend`, and also `duty_threshold -> status_threshold`, hope that's alright~~ renamed to `status.motor_utilization_threshold`

### DIFF
--- a/src/conf/settings.xml
+++ b/src/conf/settings.xml
@@ -3294,32 +3294,34 @@ p, li { white-space: pre-wrap; }
             <cDefine>CFG_DFLT_LEDS_STATUS_SHOW_SENSORS_WHILE_RUNNING</cDefine>
             <valInt>1</valInt>
         </leds.status.show_sensors_while_running>
-        <leds.status.duty_threshold>
-            <longName>Status Duty Threshold</longName>
+        <leds.status.motor_utilization_threshold>
+            <longName>Motor Utilization Threshold</longName>
             <type>1</type>
             <transmittable>1</transmittable>
             <description>&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
 &lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
 p, li { white-space: pre-wrap; }
 &lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'Roboto'; ; font-weight:400; font-style:normal;&quot;&gt;
-&lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;Duty threshold above which the duty is shown on the status bar (in amber yellow color) instead of the battery. Note there is a 10% hysteresis, meaning if you duty shows at 20%, it will be shown until it drops below 10% (to avoid constant blinking). Values below 15% set the threshold to 15%.&lt;/p&gt;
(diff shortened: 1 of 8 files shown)


src/conf/settings.xml
```diff
@@ -3294,20 +3294,18 @@ p, li { white-space: pre-wrap; }
             <cDefine>CFG_DFLT_LEDS_STATUS_SHOW_SENSORS_WHILE_RUNNING</cDefine>
             <valInt>1</valInt>
         </leds.status.show_sensors_while_running>
-        <leds.status.duty_threshold>
-            <longName>Status Duty Threshold</longName>
+        <leds.status.status_threshold>
```
lukash: `status_threshold` is a poor name. It's already `status.abc_threshold`. We should decide how this thing should be called overall. I'd say either "Motor Saturation" or Floaty calls this "Motor Utilization", which is probably better? So then this option should be called like `leds.status.motor_utilization_threshold`.

lukash (commented):
Thanks for this.

I was actually just recently thinking I should get to add the last led blinking when battery empty. Yes please definitely do that as a separate commit, it's independent from the motor utilization change. Could even go into a separate PR.

I'll have to find some time to do a full review, how determined are you to get this into 1.3? Should be possible... At first glance the code looks reasonable...

acheronfail:
> I'll have to find some time to do a full review, how determined are you to get this into 1.3? Should be possible... At first glance the code looks reasonable...

I'm not fussed, if I want these features I can just put them on my board myself.

> I was actually just recently thinking I should get to add the last led blinking when battery empty. Yes please definitely do that as a separate commit, it's independent from the motor utilization change. Could even go into a separate PR.

Sounds good, I'll remove from this PR and create a new one.

lukash:
Rather than commenting I ended up doing a bunch of fixes myself. Wanted to push it to the PR branch but for some reason it doesn't work so I pushed it here: https://github.com/lukash/refloat/tree/acheronfail/status-current-colours

In short, fixed a bug with negative battery current (previously it was dividing two negative numbers resulting in a positive ratio, you've added a `fabsf` so I added another one for consistency with the motor current and more transparency), adjusted the two new colors, cleaned a bunch of stuff on the new option and added migration from the old option to the new, since the rename would cause all users to lose this setting. It'll only work for tunes within the UI, not XML backups, but preferable to keep the option name true to what it does. It's all in fixups meant to be squashed with the original commit.

Also rebased to resolve a conflict and I adjusted the Feature: trailer text too. Please check my changes, hopefully I didn't mess anything up. I'd also like to ask if you could test the config option migration is working. You have to go from version 1.2 for it to work though. Meaning:
1. Load 1.2
2. Set the Status Duty Threshold to a non-default value
3. Load 1.3 with these commits and see if your custom value carried over

Planning on releasing 1.3 preview 3 tomorrow morning before I leave for a weekend event, would be cool to get it in, but if you don't make it no big deal, I'll merge it later. Thanks!

acheronfail:
@lukash, I've rebased and squashed in your changes :+1: thanks for those!

I just flashed 1.2, set the value to 25%, flashed 1.3 dev (6e3794e) and yes, the value carried over, so that works! :tada: 
Also just took it for a quick spin and tested that all three colours (duty + motor + battery) are working as expected as well.

If I did the git trailer properly then this should be all good to go.

lukash:
Looking all good, thank you!

PR #83, opened 2026-04-29
Haptic alerts for footpad sensors

acheronfail:
I'm interested in your thoughts on this change. But basically it's adding haptic alerts when traveling at speed without one (or both) of the footpads being triggered. Use-cases:

1. Alerting when your footing isn't right, and you're only engaging a single footpad sensor instead of two
2. Replacement for the buzzer alert right before a full fault (if your board doesn't have a buzzer)
3. For users testing new footpads, they can:
    - Enable "disable moving faults"
    - Enable this new feature
    - They can ride without fear of unexpected stops, but get haptic feedback indicating that their sensors aren't working correctly

For me, I'm more interested in the first use-case, since I have a board where one sensor only works in certain parts of the footpad, and I want alerts on it.

Some things for alignment:

- [ ] take a look at both commits - I wasn't sure if I should use the alert tracker or not, so you choose which one you prefer and I'll amend or remove the commit to include that
- [ ] do you think we should have alerts that play at different speeds?
    - i.e., they all beep at the same rate, it's only distinguishable by the frequency (pitch) of the alert. Perhaps we could have different speeds (quick for fw fault, slow for adc?) or patterns of some kind?


lukash:
I am planning on revamping the alerting altogether in 1.4 (via the Alert tracker) with the options to use beeper or haptic for each individual alert. Sensor alerts will be part of it.

I don't think we can get this into 1.3, I don't want to inflate it too much. I'll have a closer look at your implementation later but my plan is to build the new alerting from the bottom up in a systematic way and not sure how your implementation will fit...

acheronfail:
Sweet - don't worry about this change then, if it's coming soon anyway that's not an issue.

I'll just run this build on my one board with the not-so-sensitive foot sensors until 1.4 is out!

PR #84, opened 2026-05-08
On low battery pulse a single red led

acheronfail:
Feature: A pulsing red led now indicates when the board has low battery

See discussion in: https://github.com/lukash/refloat/pull/82#pullrequestreview-4251576491

### DIFF
--- a/src/leds.c
+++ b/src/leds.c
@@ -460,6 +460,20 @@ static void anim_progress_bar(
     }
 }
 
+static void anim_battery_bar(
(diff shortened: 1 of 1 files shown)


lukash:
Looks good, may I just ask to rephrase the changelog trailer?
> Feature: Flash the last remaining LED on the status battery bar if battery percentage goes below that point

acheronfail:
Done, commit amended with the updated trailer. :+1:

lukash:
Thanks!

PR #85, opened 2026-05-20
Updated RT-TUNE and TUNE-OTHER commands

surfdado:
This will allow features like negative variable tiltback as well as disable moving faults to be set via COMMAND_TUNE_OTHER. It also has some extensions to COMMAND_RT_TUNE (see individual commits)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1727,7 +1727,12 @@ static void cmd_runtime_tune_other(Data *d, unsigned char *cfg, int len) {
         if (tiltspeed > 0) {
             d->float_conf.noseangling_speed = tiltspeed / 10;
         }
-        d->float_conf.tiltback_variable = tiltvarrate / 100;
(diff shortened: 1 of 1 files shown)


lukash:
I'll close this. 1.3 beta is around the corner so hopefully no need to get this into 1.2, but the commits are in `main` now so I can backport them from there if needed.

surfdado:
> I'll close this. 1.3 beta is around the corner so hopefully no need to get this into 1.2, but the commits are in `main` now so I can backport them from there if needed.

The changes do apply cleanly on 1.2 as well, but I'll leave it up to you. Beta still means it's only available to people who install manually, whereas 1.2 would be available to everyone pretty much immediately. I can just provide 1.2 with my changes manually, to anyone who wants 1.2 but with the enhanced FC-support.

PR #86, opened 2026-05-20
Updated RT-TUNE and TUNE-OTHER commands (v1.3)

surfdado:
This will allow features like negative variable tiltback as well as disable moving faults to be set via COMMAND_TUNE_OTHER. It also has some extensions to COMMAND_RT_TUNE (see individual commits)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1722,6 +1722,9 @@ static void cmd_runtime_tune_tilt(Data *d, unsigned char *cfg, int len) {
     d->float_conf.tiltback_duty = (float) cfg[2] / 100.0;
     d->float_conf.tiltback_duty_angle = (float) cfg[3] / 10.0;
     d->float_conf.tiltback_duty_speed = (float) cfg[4] / 10.0;
+    if (len >= 6) {
+        d->float_conf.tiltback_speed = (float) cfg[5];
+    }
 
     beep_alert(d, 3, 0);
 }
--- a/src/main.c
+++ b/src/main.c
@@ -1576,8 +1576,12 @@ static void cmd_runtime_tune(Data *d, unsigned char *cfg, int len) {
 
         split(cfg[7], &h1, &h2);
         d->float_conf.atr_angle_limit = h1 + 5;
-        d->float_conf.atr_on_speed = (h2 & 0x3) + 3;
-        d->float_conf.atr_off_speed = (h2 >> 2) + 2;
+        if (h2 > 0) {
+            // kept for compatibility reasons
+            // wider ranges can be set with byte[18]
+            d->float_conf.atr_on_speed = (h2 & 0x3) + 3;
+            d->float_conf.atr_off_speed = (h2 >> 2) + 2;
+        }
 
         split(cfg[8], &h1, &h2);
         d->float_conf.atr_response_boost = ((float) h1) / 10 + 1;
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -1583,7 +1587,7 @@ static void cmd_runtime_tune(Data *d, unsigned char *cfg, int len) {
         split(cfg[15], &h1, &h2);
         float onspd = h1;
         float offspd = h2;
-        d->float_conf.torquetilt_on_speed = onspd / 2;
+        d->float_conf.torquetilt_on_speed = onspd + 3;
```
lukash: This changes behavior of the current functionality, so it has the potential to create a mess. You have to be careful to check for the version and if anyone else is actually using this command you're breaking it for them by doing this.

src/main.c
```diff
@@ -1747,6 +1752,18 @@ static void cmd_runtime_tune_other(Data *d, unsigned char *cfg, int len) {
...
 
+    if (len >= 15) {
+        int flags2 = cfg[14];
+        // bits 0 & 1:
+        d->float_conf.fault_moving_fault_disabled = ((flags2 & 0x1) == 0x1);
+        d->float_conf.is_footbeep_enabled = ((flags2 & 0x2) == 0x2);
+        d->switch_warn_beep_erpm = d->float_conf.is_footbeep_enabled ? 2000 : 100000;
+        // bits 2 & 3:
+        int pbmode = (flags2 >> 2) & 0x3;
+        d->motor_control.parking_brake_mode = pbmode;
```
lukash: You don't need this line, the config is propagated to `d->motor_control` via `reconfigure()`.

lukash (commented):
The PR is against `main` but based on `testing`, so it's pulling in 42 commits, can you rebase your two commits onto `main`?

lukash:
Also, the changelog trailers should be written from the point of end user, which doesn't know what the RT_TUNE and TUNE_OTHER commands are, could you rephrase the trailers to say something along the lines of "Support <option_list> to be set as a runtime tune (Float Control)"? (feel free to improve it)

surfdado:
Sorry I missed that, should be all fixed now, feature comments are adjusted too.

I also added a 3rd commit I had forgotten - adds speed based pushback to the TUNE_TILT command.

src/main.c
lukash: @surfdado I think you removed this line:
```c++
d->switch_warn_beep_erpm = d->float_conf.is_footbeep_enabled ? 2000 : 100000;
```
Istead of:
```c++
d->motor_control.parking_brake_mode = pbmode;
```

The `switch_warn_beep_erpm` line is still needed (or you need to move the relevant assignment from `configure()` to `reconfigure()`, but feel free to just leave the line, still slated for a refactor, coming in 1.4). But the `motor_control` line is not.

src/main.c
lukash: @surfdado I was referring to this change when talking about the potential breakage on discord. You mentioned something about the `len` being used to "gate" this, but I don't see it. I'm kinda fine leaving this as a dedicated FC command until there's a better way to set tunes in runtime only, just wanted to make sure there's no misunderstanding or mistake.

surfdado: this is correct - in Discord I suggested the "len" gating as a possible solution to address your concern, but hadn't actually implemented it yet. Since it sounds like you are okay with that approach I shall implement it shortly

surfdado: Okay, added and updated.

src/main.c
surfdado: ok, another update pushed

src/main.c
```diff
@@ -1623,6 +1632,20 @@ static void cmd_runtime_tune(Data *d, unsigned char *cfg, int len) {
         d->float_conf.kp2_brake = ((float) h2) / 10;
         beep_alert(d, 1, 1);
     }
+    if (len >= 18) {
```
lukash: For a safe interface this condition should be `>= 19` (to ensure item on index 18 is present) or add a separate `if` statement for the atr speeds. Otherwise LGTM.

surfdado: you're right, changed it

lukash:
Looking good now, thanks.

PR #90, opened 2026-07-14
Bound LCM protocol payloads

mjc:
Bounds LCM protocol inputs before storing or forwarding them.

## Changes

- Copy poll names within the fixed name buffer and always terminate them.
- Clamp the three brightness percentages to the schema-defined 0–100 range.
- Clear one-shot payload state for every enabled light-control request, including malformed short requests.
- Cap forwarded payloads at the fixed 64-byte storage capacity.
- Remove the unreachable disabled-module branch from the enabled-only handler.

## Behavior

Valid LCM traffic is unchanged. Oversized names and payloads stay within their fixed buffers, and a rejected short light-control request cannot leave an older payload queued for the next poll response.

I was able to cause a crash on the existing code with enough BTLE packets.

### DIFF
--- a/src/lcm.c
+++ b/src/lcm.c
@@ -66,13 +66,11 @@ void lcm_poll_request(LcmData *lcm, uint8_t *buffer, size_t len) {
 
     // Optionally pass in LCM name and version in a single string
     if (len > 0) {
-        for (size_t i = 0; i < MAX_LCM_NAME_LENGTH; i++) {
-            if (i > len || i > MAX_LCM_NAME_LENGTH - 1 || buffer[i] == '\0') {
-                lcm->name[i] = '\0';
-                break;
-            }
+        size_t name_len = min(len, sizeof(lcm->name) - 1);
+        for (size_t i = 0; i < name_len; ++i) {
             lcm->name[i] = buffer[i];
         }
+        lcm->name[name_len] = '\0';
     }
 }
 
--- a/src/lcm.c
+++ b/src/lcm.c
@@ -196,9 +196,9 @@ void lcm_light_ctrl_request(LcmData *lcm, unsigned char *cfg, int len) {
 
     int32_t idx = 0;
 
-    lcm->brightness = cfg[idx++];
-    lcm->brightness_idle = cfg[idx++];
-    lcm->status_brightness = cfg[idx++];
+    lcm->brightness = min(cfg[idx++], 100u);
+    lcm->brightness_idle = min(cfg[idx++], 100u);
+    lcm->status_brightness = min(cfg[idx++], 100u);
 
     if (len > 3) {
         if (lcm->enabled) {
--- a/src/lcm.c
+++ b/src/lcm.c
@@ -194,6 +194,7 @@ void lcm_light_ctrl_request(LcmData *lcm, unsigned char *cfg, int len) {
         return;
     }
 
+    lcm->payload_size = 0;
     int32_t idx = 0;
 
     lcm->brightness = min(cfg[idx++], 100u);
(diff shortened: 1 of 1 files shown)


src/lcm.c
```diff
@@ -198,23 +196,15 @@ void lcm_light_ctrl_request(LcmData *lcm, unsigned char *cfg, int len) {
 
     int32_t idx = 0;
 
-    lcm->brightness = cfg[idx++];
-    lcm->brightness_idle = cfg[idx++];
-    lcm->status_brightness = cfg[idx++];
+    lcm->brightness = min(cfg[idx++], 100u);
+    lcm->brightness_idle = min(cfg[idx++], 100u);
+    lcm->status_brightness = min(cfg[idx++], 100u);
+    lcm->payload_size = 0;
```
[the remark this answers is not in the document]
mjc: ope. I didn't interpret the code as working like that. will double check.

src/lcm.c
```diff
@@ -66,13 +66,11 @@ void lcm_poll_request(LcmData *lcm, uint8_t *buffer, size_t len) {
 
     // Optionally pass in LCM name and version in a single string
     if (len > 0) {
-        for (size_t i = 0; i < MAX_LCM_NAME_LENGTH; i++) {
-            if (i > len || i > MAX_LCM_NAME_LENGTH - 1 || buffer[i] == '\0') {
-                lcm->name[i] = '\0';
-                break;
-            }
+        size_t i = 0;
```
lukash: Since `i` is not used outside of the `for`, can you put the declaration insdie the ()?

lukash (commented):
I think the changes in the two commits in this PR should ideally be split into 3 commits:
- the simplification of the name copying code
- limiting the brightnesses to max 100
- handling the payload size overflow / clearing on short request

src/lcm.c
mjc: `i` is currently used on line 73 to add the null terminator. but I can compute the bounded length first and achieve the same thing, that's a better solution than what I was doing, I'll do that.

src/lcm.c
```diff
@@ -66,13 +66,11 @@ void lcm_poll_request(LcmData *lcm, uint8_t *buffer, size_t len) {
...
-        for (size_t i = 0; i < MAX_LCM_NAME_LENGTH; i++) {
-            if (i > len || i > MAX_LCM_NAME_LENGTH - 1 || buffer[i] == '\0') {
-                lcm->name[i] = '\0';
-                break;
-            }
+        size_t name_len = min(len, sizeof(lcm->name) - 1);
+        for (size_t i = 0; i < name_len; ++i) {
             lcm->name[i] = buffer[i];
         }
+        lcm->name[name_len] = '\0';
```
lukash, line 73: Minor: This commit actually also fixes a potential OOB read of the name, would be good to mention that in the commit message.

src/lcm.c
```diff
@@ -190,7 +190,12 @@ void lcm_get_battery_response(const LcmData *lcm) {
 }
 
 void lcm_light_ctrl_request(LcmData *lcm, unsigned char *cfg, int len) {
-    if (!lcm->enabled || len < 3) {
+    if (!lcm->enabled) {
+        return;
+    }
+
+    lcm->payload_size = 0;
+    if (len < 3) {
```
lukash: Having a second look the `payload_size` should not be cleared before the `len < 3` check, as that check is effectively for malformed request. Only a valid request should clear the payload.

lukash (commented):
A couple more remarks...

Also, applies to all the PRs, please don't use the conventional commit messages (prefixed with fix: ). I prefer component prefixes is applicable. And after the prefix, please start with a capital letter just like you do if there's no prefix (look at other commits messages if unclear). I won't force you to rephrase the current commit messages if you don't want to (just let me know so I don't wait).

src/lcm.c
```diff
@@ -190,29 +190,25 @@ void lcm_get_battery_response(const LcmData *lcm) {
...
 
 void lcm_light_ctrl_request(LcmData *lcm, unsigned char *cfg, int len) {
-    if (!lcm->enabled || len < 3) {
+    if (!lcm->enabled) {
+        return;
+    }
+
+    if (len < 3) {
         return;
     }
```
lukash, line 195: Well, now you can combine the two conditions as they were before 😁 no need for two separate `if`s.

lukash:
Thank you.

PR #91, opened 2026-07-14
Pass Konami sequence element counts

mjc:
Passes element counts, rather than byte sizes, when initializing the three footpad gesture sequences.

## Why

`konami_init` expects a `uint8_t` element count. The old code passed `sizeof(sequence)`, which is the array byte size. After a valid gesture was consumed, the state machine could therefore continue indexing past the end of the sequence.

## Changes

- Pass the element count for the flywheel, headlights-on, and headlights-off sequences.
- Convert each derived count explicitly at the `uint8_t` API boundary.
- Assert at compile time that every fixed sequence fits that representation.

## Scope

Only initialization of the three fixed gestures changes. Their contents, timing, and valid state-machine behavior are otherwise unchanged.

Hilariously, it seems possible to do a TAS-style physical reproduction of this but there is no way I am going to spend the time to try. I did get strange behavior after enough trying to trigger it manually though.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1216,16 +1216,18 @@ static void data_init(Data *d) {
 
     data_recorder_init(&d->data_record, imu_sample_rate);
 
-    konami_init(&d->flywheel_konami, flywheel_konami_sequence, sizeof(flywheel_konami_sequence));
+    konami_init(
+        &d->flywheel_konami, flywheel_konami_sequence, array_size(flywheel_konami_sequence)
+    );
     konami_init(
         &d->headlights_on_konami,
         headlights_on_konami_sequence,
-        sizeof(headlights_on_konami_sequence)
+        array_size(headlights_on_konami_sequence)
     );
     konami_init(
         &d->headlights_off_konami,
         headlights_off_konami_sequence,
-        sizeof(headlights_off_konami_sequence)
+        array_size(headlights_off_konami_sequence)
     );
 
     ema_init(&d->balance_current);
--- a/src/lib/utils.h
+++ b/src/lib/utils.h
@@ -38,6 +38,7 @@
 #define TORQUE_CONSTANT_COMPAT (1.5f * 15 * 0.027f)
 
 #define unused(x) (void) (x)
+#define array_size(array) (sizeof(array) / sizeof((array)[0]))
 
 #if defined(__GNUC__) && __GNUC__ < 9
 


src/main.c
```diff
@@ -1216,16 +1216,20 @@ static void data_init(Data *d) {
 
     data_recorder_init(&d->data_record, imu_sample_rate);
 
-    konami_init(&d->flywheel_konami, flywheel_konami_sequence, sizeof(flywheel_konami_sequence));
+    konami_init(
+        &d->flywheel_konami,
+        flywheel_konami_sequence,
+        sizeof(flywheel_konami_sequence) / sizeof(flywheel_konami_sequence[0])
```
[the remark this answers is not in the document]
mjc: I was going for minimum fix but ok copilot you win.

src/main.c
```diff
@@ -94,6 +94,15 @@ static const FootpadSensorState headlights_off_konami_sequence[] = {
...
 };
 
+_Static_assert(
+    sizeof(flywheel_konami_sequence) / sizeof(flywheel_konami_sequence[0]) <= UINT8_MAX &&
+        sizeof(headlights_on_konami_sequence) / sizeof(headlights_on_konami_sequence[0]) <=
+            UINT8_MAX &&
+        sizeof(headlights_off_konami_sequence) / sizeof(headlights_off_konami_sequence[0]) <=
+            UINT8_MAX,
+    "Konami sequence length exceeds uint8_t"
+);
```
lukash: This static assert won't scale - it requires each sequence to be added. OTOH konami sequences of more than 255 items are not expected and worst case is they won't work. So just drop this assert please.

src/main.c
```diff
@@ -1219,17 +1228,17 @@ static void data_init(Data *d) {
     konami_init(
         &d->flywheel_konami,
         flywheel_konami_sequence,
-        sizeof(flywheel_konami_sequence) / sizeof(flywheel_konami_sequence[0])
+        (uint8_t) (sizeof(flywheel_konami_sequence) / sizeof(flywheel_konami_sequence[0]))
```
lukash: The cast is not adding value (besides if we had the particular warning for this turned on, in which case the cast can be added), so please remove it too.

src/main.c
lukash: Could you add an `array_size()` macro to `lib/utils.h` and use it in the three places?

mjc:
Added array_size helper and addressed the other feedback.

mjc:
second force push was just formatting.

lukash:
Thanks!

PR #92, opened 2026-07-15
Suppress placeholder BMS faults during startup grace

mjc:
Prevent the startup BMS message-age sentinel from falling through to telemetry threshold checks during the existing grace period.

While BMS data is stale, `bms_update()` now returns without reporting telemetry faults until the grace period expires. Afterward it reports the existing connection fault; fresh telemetry continues through the existing threshold checks unchanged.

### DIFF
--- a/src/bms.c
+++ b/src/bms.c
@@ -44,9 +44,11 @@ void bms_update(BMS *bms, const CfgBMS *cfg, const Time *time) {
     const float timeout = 5.0f;
 
     // Before the first BMS update occurs right after startup, msg_age has its
-    // init value. We need to wait the `timeout` time before issuing errors.
-    if (bms->msg_age > timeout && time_elapsed(time, start, timeout)) {
-        set_fault(&fault_mask, BMSF_CONNECTION);
+    // init value. Suppress all telemetry faults until the grace period expires.
+    if (bms->msg_age > timeout) {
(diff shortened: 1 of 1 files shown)


mjc:
code review made me unsure about this and then I thought about it. minimal fix posted instead. sorry about the wild ride lmao.

lukash:
Which case is your code fixing? Seems to me they result in identical behavior because at startup `bms->fault_mask = BMSF_NONE;`.

mjc:
before:

```
bms->fault_mask = BMSF_NONE; // during init

// First bms_update():
fault_mask = BMSF_NONE;
if (42 > 5 && false) { ... } // does not return
if (0.0f < 2.7f) {
    set_fault(&fault_mask, BMSF_CELL_UNDER_VOLTAGE);
}
bms->fault_mask = fault_mask; // now contains false undervoltage
```

17d3c6104097c9275d53a9b4fae142f1eb96d08b was supposed to fix this but didn't exactly do it right.

To clarify the scope: with a healthy BMS this is likely only present until the BMS Lisp thread makes its first ext-bms call, which normally happens almost immediately. It persists through the startup grace period only when no BMS snapshot is forwarded. So this is primarily a correctness fix for the intended grace period.

lukash:
Thanks for the clarification!

PR #93, opened 2026-07-15
Preserve circular buffer item semantics

mjc:
Fix two independent circular-buffer behaviors.

`circular_buffer_iterate()` previously used an item index as a byte offset. For buffers containing multi-byte records, callbacks after the first could therefore receive pointers inside another record. The index is now multiplied by `item_size` so each callback receives a complete item. This fixes the experimental recorder plotting path, which iterates over `Sample` records.

Following review feedback, `circular_buffer_pop()` now implements a conventional FIFO pop: its unused index argument was removed, it copies the item at the tail, and then advances the tail. There are currently no callers requiring migration.

The defensive argument guards and indexed-item shifting from earlier revisions were removed following review feedback.

### DIFF
--- a/src/lib/circular_buffer.c
+++ b/src/lib/circular_buffer.c
@@ -96,7 +96,7 @@ void circular_buffer_iterate(
 
     size_t i = cb->tail;
     do {
-        callback(&cb->buffer[i], data);
+        callback(cb->buffer + i * cb->item_size, data);
         increment(cb, &i);
     } while (i != cb->head);
 }
--- a/src/lib/circular_buffer.c
+++ b/src/lib/circular_buffer.c
@@ -74,8 +74,8 @@ bool circular_buffer_get(const CircularBuffer *cb, size_t i, void *item) {
     return true;
 }
 
-bool circular_buffer_pop(CircularBuffer *cb, size_t i, void *item) {
-    if (!circular_buffer_get(cb, i, item)) {
+bool circular_buffer_pop(CircularBuffer *cb, void *item) {
+    if (!circular_buffer_get(cb, 0, item)) {
         return false;
     }
 
--- a/src/lib/circular_buffer.h
+++ b/src/lib/circular_buffer.h
@@ -53,11 +53,10 @@ void circular_buffer_push(CircularBuffer *cb, const void *item);
 bool circular_buffer_get(const CircularBuffer *cb, size_t i, void *item);
 
 /**
- * Pops an item at index i (a position relative to the current tail) from the
- * buffer. If there's no item at that index, leaves item unchanged and returns
- * false, otherwise returns true.
+ * Pops the item at the tail of the buffer. If the buffer is empty, leaves item
+ * unchanged and returns false, otherwise returns true.
  */
-bool circular_buffer_pop(CircularBuffer *cb, size_t i, void *item);
+bool circular_buffer_pop(CircularBuffer *cb, void *item);
 
 /**
  * Iterates over the buffer, calling the callback function on each item.


src/lib/circular_buffer.c
```diff
@@ -96,7 +107,7 @@ void circular_buffer_iterate(
 
     size_t i = cb->tail;
     do {
-        callback(&cb->buffer[i], data);
+        callback(cb->buffer + i * cb->item_size, data);
```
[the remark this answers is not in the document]
mjc: ope. fixing that. C is so tricky sometimes.

src/lib/circular_buffer.c
```diff
@@ -50,6 +50,10 @@ size_t circular_buffer_size(const CircularBuffer *cb) {
 }
 
 void circular_buffer_push(CircularBuffer *cb, const void *item) {
+    if (cb->length == 0 || cb->item_size == 0 || cb->buffer == NULL || item == NULL) {
+        return;
+    }
```
lukash: These are effectively guarding against programmer error passing invalid data or not initializing the buffer correctly. I don't see much difference to passing a null pointer as `cb` to any of the functions or passing other invalid values elsewhere, OTOH it's adding unnecessary checks to each call. I'd prefer to not add this.

src/lib/circular_buffer.c
```diff
@@ -79,6 +84,12 @@ bool circular_buffer_pop(CircularBuffer *cb, size_t i, void *item) {
         return false;
     }
 
+    while (i > 0) {
+        size_t dst = (cb->tail + i) % cb->length;
+        size_t src = (cb->tail + i - 1) % cb->length;
+        memcpy(cb->buffer + dst * cb->item_size, cb->buffer + src * cb->item_size, cb->item_size);
+        --i;
+    }
```
lukash: What is this supposed to be doing?

src/lib/circular_buffer.c
mjc: removed the guards

src/lib/circular_buffer.c
mjc: `circular_buffer_pop()` accepts an index.  previously it returned the item at that index while always removing the item at the tail. This loop shifts the preceding items forward so advancing the tail removes the requested item instead. For example, popping index 1 from [A, B, C] returns B and leaves [A, C].

Worth noting that there's currently no callers of this - but if it were to be called, it would not do what the docs said, thus leaving a hard to find bug.

Alternatively, deleting this would probably work just fine. It’s a footgun if left as-is for someone to use later.

lukash: Ok, I didn't even notice the `i` there when reading the PR - the pop method was written for completeness and I honestly don't know how it came around with index as an argument (maybe as an analogy to the get method...). For a circular buffer queue implementation the pop should just remove from the back, the index and shifting the array doesn't really make sense. So if you want just remove the index and pop from the tail, but the shifting of elements is not desirable.

Prime unit test material for sure.

mjc: yep, I caught that when doing the test suite. I'll update this to work the way you suggested.

lukash:
Minor, the commit title "Circular buffer: Remove indexed items" doesn't make much sense, could you improve it? Besides that looks good now.

mjc:
should be all set now. sorry about the amount of churn, doing my best

lukash:
No worries, thanks for the patience, I think it's good to maintain a readable history. FWIW I prefer a bit more descriptive commit titles overall (not too much though), but this is good enough :grin:

PR #94, opened 2026-07-15
Harden LED configuration and color packing

mjc:
Validate LED strip color order before allocating DMA storage and report which strip is invalid.

The paint path also rejects unexpected encodings, and channel values are widened before bit packing. Valid GRB, GRBW, RGB, and WRGB output is unchanged.

### DIFF
--- a/src/led_driver.c
+++ b/src/led_driver.c
@@ -241,20 +241,20 @@ inline static uint8_t cgamma(uint8_t c) {
 
 static uint32_t color_grb(uint8_t w, uint8_t r, uint8_t g, uint8_t b) {
     unused(w);
-    return (g << 16) | (r << 8) | b;
+    return ((uint32_t) g << 16) | ((uint32_t) r << 8) | b;
 }
 
 static uint32_t color_grbw(uint8_t w, uint8_t r, uint8_t g, uint8_t b) {
-    return (g << 24) | (r << 16) | (b << 8) | w;
+    return ((uint32_t) g << 24) | ((uint32_t) r << 16) | ((uint32_t) b << 8) | w;
 }
 
 static uint32_t color_rgb(uint8_t w, uint8_t r, uint8_t g, uint8_t b) {
     unused(w);
-    return (r << 16) | (g << 8) | b;
+    return ((uint32_t) r << 16) | ((uint32_t) g << 8) | b;
 }
 
 static uint32_t color_wrgb(uint8_t w, uint8_t r, uint8_t g, uint8_t b) {
-    return (w << 24) | (r << 16) | (g << 8) | b;
+    return ((uint32_t) w << 24) | ((uint32_t) r << 16) | ((uint32_t) g << 8) | b;
 }
 
 void led_driver_paint(LedDriver *driver) {
--- a/src/led_driver.c
+++ b/src/led_driver.c
@@ -166,20 +166,6 @@ static void deinit_hw(const PinHwConfig *cfg) {
     disable_dma_stream(cfg);
 }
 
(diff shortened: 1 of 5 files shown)


src/led_driver.c
```diff
@@ -282,6 +294,8 @@ void led_driver_paint(LedDriver *driver) {
         case LED_COLOR_WRGB:
             color_conv = color_wrgb;
             break;
+        default:
+            return;
```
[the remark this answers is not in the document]
mjc: this seems not necessary; setup already rejects this and this might generate a lot of logs.

src/led_driver.c
```diff
@@ -241,20 +249,20 @@ inline static uint8_t cgamma(uint8_t c) {
...
 
 static uint32_t color_rgb(uint8_t w, uint8_t r, uint8_t g, uint8_t b) {
     unused(w);
-    return (r << 16) | (g << 8) | b;
+    return ((uint32_t) r << 16) | ((uint32_t) g << 8) | b;
 }
 
 static uint32_t color_wrgb(uint8_t w, uint8_t r, uint8_t g, uint8_t b) {
-    return (w << 24) | (r << 16) | (g << 8) | b;
+    return ((uint32_t) w << 24) | ((uint32_t) r << 16) | ((uint32_t) g << 8) | b;
```
lukash: These should go into a separate commit.

src/led_driver.c
```diff
@@ -282,6 +290,8 @@ void led_driver_paint(LedDriver *driver) {
         case LED_COLOR_WRGB:
             color_conv = color_wrgb;
             break;
+        default:
+            return;
```
lukash: Don't add a default to the switch, you'll silence the warning in case of an unhandled case. Instead check the function pointer is not null. But better yet, this should be validated in `Ieds_setup()`. The `color_conv` could be added as a struct member and filled in `leds_setup()` (saves some cycles on each paint), if it's invalid, log and abort the LED setup. Could do it before the color order bits and leave that out altogher, one check is enough.

src/leds.c
```diff
@@ -847,16 +847,33 @@ void leds_setup(Leds *leds, CfgHwLeds *hw_cfg, const CfgLeds *cfg) {
...
             front_offset = current_offset;
             current_offset += leds->front_strip.length;
             strip_array[strip_i++] = &leds->front_strip;
         } else if (hw_cfg->rear.order == i && hw_cfg->rear.count > 0) {
             led_strip_configure(&leds->rear_strip, &hw_cfg->rear);
+            leds->rear_strip.color_conv = led_driver_color_converter(leds->rear_strip.color_order);
+            if (!leds->rear_strip.color_conv) {
+                log_error("Invalid rear LED color order: %u.", hw_cfg->rear.color_order);
+                return;
+            }
```
lukash, line 876: This duplication is not good. It's about architecturally putting the code to the correct place. Looks like the whole color_order handling could be moved into the led_strip module, storing the bit number and the converter in the LedStrip struct. They are not really LedStrip related by nature, they should probably technically go into their own module (e.g. led_color_order.{c,h} or just make a directory for the led code at this stage). But I'd accept it if the functions were cleanly placed into led_strip.c too. More structural work and I don't really wanna spend the time iterating on this (so to be blunt if you can do it right please do, if not, let's not spend the time 😅 ).

A straightforward fix of the uninitialized pointer dereference would be to initialize it and early-return if it's NULL.

mjc: sorry I should not have pushed until I had a chance to re-review the larger context. I'll put this back to draft and get back to you when it's really right.

a bit of a struggle since there's so much to keep track of, but I'll get back to you when it feels right.

PR #95, opened 2026-07-15
Plot decoded recorder samples in seconds

mjc:
Decode stored half-precision recorder values before plotting and convert tick timestamps to seconds on the x-axis.

Recorder storage and binary data responses are unchanged.

### DIFF
--- a/src/conf/buffer.c
+++ b/src/conf/buffer.c
@@ -40,6 +40,14 @@ uint16_t to_float16(float x) {
     return (b&0x80000000)>>16 | (e>112)*((((e-112)<<10)&0x7C00)|m>>13) | ((e<113)&(e>101))*((((0x007FF000+m)>>(125-e))+1)>>1) | (e>143)*0x7FFF;
 }
 
+float from_float16(uint16_t x) {
+    const float sign = x & 0x8000u ? -1.0f : 1.0f;
+    const uint16_t exponent = (x >> 10) & 0x1fu;
+    const uint16_t mantissa = x & 0x03ffu;
+    return sign * (exponent ? ldexpf(1024.0f + mantissa, exponent - 25)
+                            : ldexpf(mantissa, -24));
+}
+
 // clang-format on
 #pragma GCC diagnostic pop
 
--- a/src/data_recorder.c
+++ b/src/data_recorder.c
@@ -134,11 +134,10 @@ static void send_point_vt_experiment(const void *item, void *data) {
     unused(data);
 
     Sample *sample = (Sample *) item;
+    const float time = sample->time * (1.0f / SYSTEM_TICK_RATE_HZ);
     for (uint8_t i = 0; i < ITEMS_COUNT_REC(RT_DATA_ALL_ITEMS); ++i) {
         VESC_IF->plot_set_graph(i);
-        VESC_IF->plot_send_points(
-            sample->time * (1.0f / SYSTEM_TICK_RATE_HZ), from_float16(sample->values[i])
-        );
+        VESC_IF->plot_send_points(time, from_float16(sample->values[i]));
     }
 }
 
--- a/src/data_recorder.c
+++ b/src/data_recorder.c
@@ -136,7 +136,9 @@ static void send_point_vt_experiment(const void *item, void *data) {
     Sample *sample = (Sample *) item;
     for (uint8_t i = 0; i < ITEMS_COUNT_REC(RT_DATA_ALL_ITEMS); ++i) {
         VESC_IF->plot_set_graph(i);
-        VESC_IF->plot_send_points(sample->time, sample->values[i]);
+        VESC_IF->plot_send_points(
+            sample->time * (1.0f / SYSTEM_TICK_RATE_HZ), from_float16(sample->values[i])
+        );
     }
 }
 
--- a/src/data_recorder.c
+++ b/src/data_recorder.c
@@ -131,10 +131,10 @@ void data_recorder_sample(DataRecord *dr, const Data *d, time_t time) {
 }
 
 static void send_point_vt_experiment(const void *item, void *data) {
-    unused(data);
+    const time_t start_time = *(const time_t *) data;
 
-    Sample *sample = (Sample *) item;
-    const float time = sample->time * (1.0f / SYSTEM_TICK_RATE_HZ);
+    const Sample *sample = (const Sample *) item;
+    const float time = (sample->time - start_time) * (1.0f / SYSTEM_TICK_RATE_HZ);
     for (uint8_t i = 0; i < ITEMS_COUNT_REC(RT_DATA_ALL_ITEMS); ++i) {
         VESC_IF->plot_set_graph(i);
         VESC_IF->plot_send_points(time, from_float16(sample->values[i]));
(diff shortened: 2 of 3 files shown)


src/conf/buffer.c
```diff
@@ -40,6 +40,14 @@ uint16_t to_float16(float x) {
     return (b&0x80000000)>>16 | (e>112)*((((e-112)<<10)&0x7C00)|m>>13) | ((e<113)&(e>101))*((((0x007FF000+m)>>(125-e))+1)>>1) | (e>143)*0x7FFF;
 }
 
+float from_float16(uint16_t x) {
+    const float sign = x & 0x8000u ? -1.0f : 1.0f;
+    const uint16_t exponent = (x >> 10) & 0x1fu;
+    const uint16_t mantissa = x & 0x03ffu;
+    return sign * (exponent ? ldexpf(1024.0f + mantissa, exponent - 25)
+                            : ldexpf(mantissa, -24));
+}
```
lukash, line 49: Where does this code come from? I'd prefer to use the form from the stack overflow post.

src/data_recorder.c
```diff
@@ -134,11 +134,10 @@ static void send_point_vt_experiment(const void *item, void *data) {
     unused(data);
 
     Sample *sample = (Sample *) item;
+    const float time = sample->time * (1.0f / SYSTEM_TICK_RATE_HZ);
     for (uint8_t i = 0; i < ITEMS_COUNT_REC(RT_DATA_ALL_ITEMS); ++i) {
         VESC_IF->plot_set_graph(i);
-        VESC_IF->plot_send_points(
-            sample->time * (1.0f / SYSTEM_TICK_RATE_HZ), from_float16(sample->values[i])
-        );
+        VESC_IF->plot_send_points(time, from_float16(sample->values[i]));
```
lukash, line 140: Amend this to the first commit.

src/conf/buffer.c
mjc: I missed that the existing encoder’s cited Stack Overflow answer also included the matching decoder. 

This seems to be a small piece of "works on my compiler" archaeology, in the original Stack Overflow answer.

The encoder from the post evaluates an invalid shift when the exponent field is at most 93 or at least 126, including both 0.0f and ordinary values such as 1.0f. The decoder does so whenever the mantissa is zero, including zero and normalized powers of two. Its sign-bit shift also needs to be made unsigned. UBSan catches these, though the current Cortex-M4 output happens to produce the intended results.

mjc: Here’s how to do both, without UB, while retaining branchless output with GCC 13.3.1 and the current Cortex-M4 -Os flags:

```c
#include <string.h>

uint16_t to_float16(float x) {
    uint32_t bits;
    memcpy(&bits, &x, sizeof(bits));

    const uint32_t b = bits + 0x00001000;
    const uint32_t e = (b & 0x7F800000) >> 23;
    const uint32_t m = b & 0x007FFFFF;

    /*
     * C's conditional operator only evaluates the selected side, so
     * (125 - e) is used as a shift count only for 102 <= e <= 112.
     */
    const uint32_t denormalized = (e < 113 && e > 101)
        ? ((((0x007FF000 + m) >> (125 - e)) + 1) >> 1)
        : 0;

    return (b & 0x80000000) >> 16 |
           (e > 112) *
               ((((e - 112) << 10) & 0x7C00) | (m >> 13)) |
           denormalized |
           (e > 143) * 0x7FFF;
}

float from_float16(uint16_t x) {
    const uint32_t e = (x & 0x7C00) >> 10;
    const uint32_t m = (x & 0x03FF) << 13;

    const float mf = (float)m;
    uint32_t mf_bits;
    memcpy(&mf_bits, &mf, sizeof(mf_bits));

    const uint32_t v = mf_bits >> 23;

    /*
     * For m == 0, this produces a safe shift count of zero.
     * For m != 0, 150 - v is already in the valid 1..10 range.
     */
    const uint32_t shift = (150 - v) * (m != 0);

    const uint32_t bits =
        ((uint32_t)(x & 0x8000) << 16) |
        (e != 0) * ((e + 112) << 23 | m) |
        ((e == 0) & (m != 0)) *
            ((v - 37) << 23 |
             ((m << shift) & 0x007FE000));

    float result;
    memcpy(&result, &bits, sizeof(result));
    return result;
}
```

to_float16 retains the same 104-byte code size, with a different instruction sequence.

from_float16 pays an extra 12 bytes for the safe zero selection. no data-dependent branches in either.

The memcpy calls are the standard defined bit-cast idiom and optimize away with both GCC and Clang.

Clang lowers this conditional to branches. Masking the shift count with & 31 keeps it branchless there too, but grows Clang’s encoder from 104 to 106 bytes and GCC’s from 104 to 108 bytes, so I’ve left the GCC-optimal form here.

original encoder asm: shift unconditionally, discard invalid results after.
revised encoder asm: predicate shift with an IT block.
decoder does `moveq` to replace the shift count with zero when the mantissa is zero.

I understand if you might not want this, hence posting it as a comment

mjc:
CodeRabbit caught that the recorder guarantees strictly increasing `time_t` timestamps, but converting absolute tick values directly to float seconds can collapse adjacent samples on longer recordings. I fixed this by subtracting the first sample’s timestamp before converting to seconds. The stored/downloaded timestamps and recording behavior are unchanged; only the plot X-axis conversion is affected.

lukash:
I don't know about this. First off, I used the experimental plot support before implementing the AppUI renderer. The experimental plot is buggy in itself (the transport is not resilient against comms issues and mangles the data) and I just kept the function to output to it in case it would be useful in the future (and it has rotted since then already). Not a priority spend time on for me atm.

Regarding the actual encoding, not sure I understood the severity of the issue, it's not described too well. You replaced both encoder and decoder with completely different functions and the whole things looks AI-ed. Not sure you've done any testing. Note there is a decoding function in AppUI and it's also cited in the docs for the package commands (not sure those are affected too or the UB is native architecture specific).

mjc:
fine by me. it's not AI written - the techniques are from the CS class I took on data layouts. I used AI to make running UBSan easier and for comparing asm outputs.

worth noting the rounding is incorrectly described in the comment in the regular version - it says it rounds to nearest even, but it does not do that.

IMO the stackoverflow code is horrifying and I would not have used it in a professional setting due to reliance on UB. memcpy is how you do type punning with valid C semantics.

lukash:
@mjc ok, apologies for calling it AI'd, in that case I guess you've invested a lot of time into the in-depth analysis, I wouldn't expect that, thank you. Re-reading it's also very well described (so apologies for that as well), I guess the problem is I don't understand it and it'd take investment from me too to figure it out. Another case where unit tests would come very useful.

Reading it again, we could merge this. The experimental plot is a bit of a toss-up, but doesn't cost much to keep for now and in that case, better to have it in a good shape. Having the decoder for the float16 in C won't hurt either.

Problem is I have a huge backlog of higher priority items, so hard to find time for this and for pushing for the unit tests to happen in a future-proof manner. But for that I'm slowly paving the way, I just got the upstream `bldc` GCC updated to 14, which is also the compiler used for the packages (on the VESC build server). That's a bit orthogonal but means the ancient GCC can't become a blocker or complicate stuff mid-way some migration. Gonna try for CMake next.

PR #96, opened 2026-07-15
Honor throttle inversion for command remote input

mjc:
Apply configured throttle inversion to command-driven remote input before storing the normalized value and deriving the wheel-move speed target.

Hardware and UART remote input, timeouts, and speed limits are unchanged.

I was unsure if this is something we want but squinting at the code and UI settings made it seem like this is something people should be able to do with remotes.

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -121,11 +121,11 @@ void remote_input(Remote *remote, const Time *time, const RefloatConfig *config)
 void remote_command_input(
     Remote *remote, float value, const Time *time, const RefloatConfig *config
 ) {
-    remote->input = value;
+    remote->input = config->inputtilt_invert_throttle ? -value : value;
     if (time_elapsed(time, disengage, 2.0f)) {
         // default to a limit of 5 km/h if limit of 0 is configured for the remote
         float speed_max = config->remote.max_move_speed > 0 ? config->remote.max_move_speed : 5;
-        remote->move_speed = value * speed_max;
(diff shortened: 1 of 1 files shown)


src/remote.c
```diff
@@ -121,11 +121,11 @@ void remote_input(Remote *remote, const Time *time, const RefloatConfig *config)
 void remote_command_input(
     Remote *remote, float value, const Time *time, const RefloatConfig *config
 ) {
-    remote->input = value;
+    remote->input = config->inputtilt_invert_throttle ? -value : value;
     if (time_elapsed(time, disengage, 2.0f)) {
         // default to a limit of 5 km/h if limit of 0 is configured for the remote
         float speed_max = config->remote.max_move_speed > 0 ? config->remote.max_move_speed : 5;
-        remote->move_speed = value * speed_max;
+        remote->move_speed = remote->input * speed_max;
```
lukash, line 128: Pretty sure this is as intended. You're changing behavior without any justification. I've made recent changes to this, unless I messed up (refer to the exact change if you look it up), it's been working like this for years and should stay the way it is.

mjc: You’re right. `d505d824` introduced the unified `REMOTE` command, and I missed that its value is already normalized app control rather than raw physical-remote input. The physical path has also intentionally applied inversion only to tilt—not wheel movement—since `344c458e`. This PR changes both behaviors, so I’m closing it.

PR #97, opened 2026-08-12
Ghost Buster / Configurable Reverse Stop

pbz:
This pull request is definitely not ready to merge yet, at all, but worked on this for my own use and wanted to see if the Refloat devs were interested, but also to make these changes available for others who might need them.

Anyway, my motivation for this was getting tire of fixing my board after it ghosted and rammed itself into things. It occurred to me that reverse stop was an easy solution to this for the typical tail heavy board, but I like being able to go in reverse, at least a little. 

So this pull request just makes reverse stop a bit more configurable, and introduces one tiny change to the logic, you only have to drop below the trigger ERPM to stop accumulating towards the cutoff threshold rather than reverse direction. 

I found the following values suitable for a ghost prone board, outdoors to allow enough reversing with gentle warning before the tail comes down completely: 
  Reverse Stop Tolerance: 1,000,000 ERPM
  Reverse Stop Trigger: 1,500 ERPM
  Reverse Stop ERPM Increment: 0.00000800

If the devs are interested, I'd be happy to develop this on top of the reworked reverse stop that's on main, before/after that's out of beta.


lukash:
Hello, this anti-ghosting measure should not be implemented through Reverse Stop at all. The main point of Reverse Stop is to gently put the tail down after a certain distance. For anti-ghosting, you want to ideally give a haptic warning once starting to ride backwards and just shut the motor after some distance / speed.

I'd much prefer this was discussed on discord prior to implementation. The PR is against 1.2 and the Reverse Stop was re-written, and the PR is kind of all over the place, it's not really useful.

PR #98, opened 2026-09-11
Initialize the beeper before setting up LEDs

Michael-ango:
LEDs configured on pin B6 stopped working for me after upgrading from 1.2.1.

`leds_setup()` configures B6 as `PAL_MODE_ALTERNATE(2)` so TIM4 CH1 drives the pad.
`beeper_init()` then calls `io_set_mode(VESC_PIN_PPM, VESC_PIN_MODE_OUTPUT)` on the
same pad, which switches MODER to general-purpose output and disconnects the
alternate-function mux. The timer and DMA keep running and `led_driver_paint()` keeps
writing frames, but nothing reaches the pin, and nothing is logged.

Confirmed `VESC_PIN_PPM` and `LED_PIN_B6` are the same pad: with the pin held low,
evaluating `(gpio-configure 'pin-ppm 'pin-mode-in-pu)` released it to 5 V.

Note that disabling the beeper doesn't avoid this, since the guard is an OR, it runs
regardless unless `inputtilt_remote_type` happens to be PPM.

Tested: built unmodified from source and reproduced the fault, then confirmed this patch
fixes it on hardware.


Michael-ango:
Additionally, a check could be implemented that checks if the LEDs and beeper are configured on the same pin and output a log_error() to avoid silent failures like this in the future. The reordering I'm proposing would flip the conflict silence if a user accidentally configures both the beeper and led at the same time. The beeper would fail with no clear reason why. 

Currently this commit is the simplest approach to clearing the unavoidable conflict, an additional check is just a quality of life improvement. A beeper/led conflict after this would require negligent configuration.

PR #99, opened 2026-09-14
Make LEDs more intuitive

OakCityDirtyRandy:
LEDs are fun and people who aren't mechanical engineers can play with them without wondering if they're putting their safety at risk.  Unfortunately, the user-facing namespace and some elements of the UI can be confusing.  There are also some LED gotchas that can lead to unexpected results, and these might make some people wonder if something is wrong with the board.  

These are mostly changes to settings.xml to try and help make the experience easier.  A minor change to the Refloat tab has also been made to use the same naming conventions for the lighting on/off buttons.  No changes to enums, variables or any functional aspects under the hood have been made to ensure things work smoothly and to provide backwards compatibility.  (If these changes are adopted it might make sense to make changes to the underlying namespace for the sake of good coding hygiene.) 

Here is an overview of the changes:

- **Specs has been moved before LEDs.** Specs are more of a hardware configuration menu, whereas LEDs are a runtime menu.  It seems more intuitive to turn handle hardware-level stuff first.
- **The terms Headlights and LEDs are renamed Riding LEDs and Idle LEDs.** In Refloat, sometimes LEDs mean all lighting.  Sometimes they mean just the LEDs that are operating when the board is idle.  Headlights and taillights never cease being LEDs in the literal sense, being a headlight or taillight depends on direction of travel rather than some immutable condition, and what are currently called Headlights and Taillights can be configured to behave like what are currently called LEDs and vice versa.    _The real difference is that you're either configuring the behavior of LEDs while riding or while idle, and headlights versus taillights depends on the direction of travel._ 
- **"Turning on" Riding LEDs and Idle LEDs is now under Specs.** What is currently called _turning on_ LEDs (meaning LED behavior while idle) and Headlights (meaning LED behavior while riding) might more intuitively be called _enablement_.  These settings usually require a board restart, which makes them more similar to other params in Specs than those under LEDs, which are runtime configurable.  Specs should be set-it-and-forget-it.  Actually turning lights on and off should be done in the Refloat tab.
- **Status bar stuff has been lightly tweaked for clarity.**
- **Various little gotchas have been documented in the Help.** These concern things like the side effects of misconfiguring strip length in a daisy-chained setuo or setting Brightness high in setups without a dedicated W channel.

I hope these changes seem worthwhile.


OakCityDirtyRandy:
Good discussion [here](https://discord.com/channels/1346864089170710739/1552322462346383370).  I'm pulling this PR and will send a few smaller, more modest changes that hopefully help people navigate LEDs.

